PR #13, opened 2013-04-23
Avoid updating contexts on route change unless necessary

pdf:
On route change, contexts for dynamic segments are always marked updated,
even when identical. Additionally, any segment after the first dynamic
segment is forced to deserialize, even if the dynamic segments have not
changed.

This change tracks whether any preceeding segments have changed before 
updating contexts, and avoids unnecessary reloads/redraws.

I mistakenly submitted this against the vendored version in Ember first, because I wasn't familiar with the project structure, so some of you may have seen this before.

I'm afraid I'm having trouble coming up with tests for this (I tried from a few different angles), because the current behaviour doesn't appear to be tested.  It should be pretty obvious what this is trying to achieve, and for routes with multiple or deeply nested dynamic segments, triggering a reload of every segment after the first dynamic segment is really harsh, so I'd love to get this merged, but I need a little help.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -156,6 +156,7 @@ Router.prototype = {
         toSetup = [],
         startIdx = handlers.length,
         objectsToMatch = objects.length,
+        parentObjectChanged = false,
         object, objectChanged, handlerObj, handler, names, i;
 
     // Find out which handler to start matching at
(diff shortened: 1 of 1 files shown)


sandstrom:
@pdf Can this issue be closed? It's quite old, so I'm guessing it's outdated anyway (and if not, it would be great if the conflicts could be solved).

pdf:
I no longer use ember/router.js, and given the significant divergence in the codebase I can't quickly tell if this still applies. I'd hope not, since this was a serious performance issue - I'll just close it out.

sandstrom:
@pdf Understand, thanks for the update!

PR #56, opened 2013-10-21
Prevent error-swallowing within route hooks

machty:
Example: http://jsbin.com/ucanam/1556/edit

This depends on https://github.com/tildeio/rsvp.js/pull/124


sandstrom:
@machty This is a three year old PR. Can it be closed, or is this still an issue?

Haven't read the PR, but judging by the title my guess is that the the [error action](https://guides.emberjs.com/v2.5.0/routing/loading-and-error-substates/) can be used instead.

nathanhammond:
So I think this would be pretty neat (as a power user). That the promise chain wraps the entire function instead of just the return value though is probably a consequence of _not_ forcing the user to return a promise which you described as an implementation detail.

I can already imagine how much different the experience would be (nicer, in my opinion) but we're now three years removed from taking no action on this. At this point I feel like the behavior is so ingrained that I don't believe we could make the switch.

We can adopt this behavior if/when we move something like [`ember-prefetch`](https://github.com/nickiaconis/ember-prefetch/) into core.

Thoughts, @machty?

wagenet:
Closing due to staleness.

PR #90, opened 2014-04-10
Transition.then accepts the label parameter

ppcano:
Now, `then` accepts the same parameters than the `promise.then` implementation and passes all the arguments to the internal promise instance.

### DIFF
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -113,9 +113,10 @@ Transition.prototype = {
 
     @param {Function} success
     @param {Function} failure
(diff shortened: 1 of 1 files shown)


sandstrom:
@ppcano Hey! Just stumbled on this old issue. Is this something that can be closed? Otherwise, an update to avoid the conflict would be awesome ⛵️

rwjblue:
Looks like this landed separately: https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L167

Closing...

PR #91, opened 2014-04-16
'_id' variable name issue, increase MongoDB compatibility

canyaman:
'_id' variable name which is used in model can be placed in the route path
increase MongoDB compatibility
'post_id' regEx should be ".+_id$" neither ".*_id$" nor "_id$"


sandstrom:
@canyaman can this issue be closed?

Your problem is easily solved with the Ember Data Serializer, so no need to change how the router works depending on the backend-end id formatting.

Issue #100, opened 2014-04-24
Unable to use router.js standalone


boxinator:
Ive created a simple html page which includes rsvp, route-recognizer, and router.js, and setup a new router object as seen in the usage docs. I get the following error when instantiating a new Router object:

TypeError: object is not a function

Router() seems to return an object with a 'default' property, which seems to have the expected function to init a Router object.

Is router.js usable in a standalone fashion, or is Ember a requirement?

PR #101, opened 2014-04-24
Fix bug in build of router.js

pangratz:
The `default` export needs to be used when requiring the `router` module.

This addresses #100


nathanhammond:
This PR has effectively expired. Thanks for trying to help, sorry we didn't get back to it!

PR #124, opened 2014-07-27
Added documentation

marcelerz:
When reading through the source-code, I made some notes that resulted in this documentation here. I think that this might help also other programmers to understand the code more easily.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -5,23 +5,69 @@ export default Router;
 /**
  * Handler object
  *
- * @property {} context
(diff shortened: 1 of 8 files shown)


sandstrom:
@marcelerz Is this still relevant? (it's an old issue). If so it would be great if the conflicts could be solved, and otherwise perhaps it can be closed?

rwjblue:
I'm going to go ahead and close this, I am very sorry that we let it linger so long. 😢

PR #125, opened 2014-07-28
Add missing queryParams to three NamedTransitionIntent instantiations

marcelerz:
In some cases, the queryParams are never set even though they might be available. In two of three cases, this is not too bad since the assignment is done later on in the process, but they are generally missing in the applyToState() call, possibly creating bugs in the future when applyToState requires the queryParams to be set.
In one case, however, the queryParams are missing and are never set afterwards. I fixed it and extended one of the tests to test for it.

### DIFF
--- a/lib/router/router.js
+++ b/lib/router/router.js
@@ -263,7 +263,7 @@ Router.prototype = {
 
     // Construct a TransitionIntent with the provided params
     // and apply it to the present state of the router.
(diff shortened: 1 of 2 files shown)


sandstrom:

rwjblue:
Looks like this landed in another PR / bugfix: https://github.com/tildeio/router.js/blob/master/lib/router/router.js#L253

Thank you again for your time to submit a fix, and I am sorry for the massive delay / lack of response.

PR #126, opened 2014-07-28
Fix for incorrect handlerWithError parameter for rejected promise during redirect

marcelerz:
When throwing an error in a redirect function of a handler, the error event reports "handlerWithError" as the next handler in the handlerInfos, not the current handler. This is caused by incrementing payload.resolveIndex before redirect() is called. 
However, moving the incrementation after the redirect call introduces other problems since the redirect body expects the resolution of the handler to be completed. (For example the test "The router: transitionTo will soak up resolved all models of active transition, including present route's resolved model" would fail).
This fix is not really elegant, but it will fix this issue with minimal impact to other code.

### DIFF
--- a/lib/router/transition-state.js
+++ b/lib/router/transition-state.js
@@ -38,6 +38,7 @@ TransitionState.prototype = {
 
     var currentState = this;
     var wasAborted = false;
+    var redirect = false;
 
     // The prelude RSVP.resolve() asyncs us into the promise land.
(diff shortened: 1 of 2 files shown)


sandstrom:

rwjblue:
Sorry for the massive delay here. Closing for now (it is likely that a new PR would be needed due to rebase issues anyways).

PR #147, opened 2015-02-11
Keep dist up-to-date

Dremora:
I'm not sure what the policy is in regards to running grunt and keeping dist folder up-to-date (judging on previous commits, it seems to be quite sporadic). We have started using router.js in our project, and there's a bugfix we rely upon for which the code hasn't been built yet.

Of course, even better, it would be nice to see router.js as a bower or npm package.


nathanhammond:

We'll release new versions as seems to make sense. Thanks for your PR, and sorry to let it languish for so long!

PR #70, opened 2013-12-17
Prepare for publishing to NPM

spikebrehm:
This PR is one step in addressing tildeio/router.js#69

~~It depends on https://github.com/tildeio/route-recognizer/pull/16; once this is merged in, I'll update the ref in package.json to point to the NPM module instead of the git repo.~~
- Update `package.json`:
  - Add `main` field, pointing to CommonJS build at `dist/commonjs/main.js`.
  - Add `route-recognizer` and `rsvp` as dependencies.
  - Change `name` field from `router` to `router.js`, which is unique in NPM and allows using Browserify with `require('router.js');`.
- Add `.npmignore`, ignoring build tasks.

PR #149, opened 2015-02-25
Prepare for publishing to NPM (Redux)

pzuraq:
Adds RSVP and RouteRecognizer as package dependencies, makes all import statements either relative paths or imports from the mentioned packages. Sets the `main.js` to the CommonJS distribution for easy importing.

Continued from #70, thanks to @spikebrehm and @cheapsteak, along with everyone else!


spikebrehm:
Nice :+1:

pzuraq:
Ok, there are some problems with actually using this package in the wild. For some reason the compiled CJS file is expecting a `default` property to be exported by `RouteRecognizer`, but it doesn't actually export one:

```
// dist/commonjs/router/router.js
var RouteRecognizer = require("route-recognizer")["default"];
```

```
// route-recognizer/dist/route-recognizer.js
    if (typeof define === 'function' && define['amd']) {
      define(function() { return $$route$recognizer$$default; });
    } else if (typeof module !== 'undefined' && module['exports']) {
      module['exports'] = $$route$recognizer$$default;
    } else if (typeof this !== 'undefined') {
      this['RouteRecognizer'] = $$route$recognizer$$default;
    }
```

I'm still working on why this is happening, I think it may have something to do with the ES6 version of `RouteRecognizer`.

pzuraq:
Superceded by #167

PR #155, opened 2015-03-20
Fix UnrecognizedURLError inheritance

ef4:
This already has appropriate test coverage, which was appropriately failing.


ef4:
Sanity checked this with @stefanpenner, so I'm merging.

stefanpenner:
:+1: sorry, this was my fault.

PR #161, opened 2015-06-22
handler-info: paramsMatch compares number of properties

Aintaer:
paramsMatch had a note that explained the assumption that incoming
params would have same number of keys. Rather than assume that, test
that they actually have the same number of keys.

### DIFF
--- a/lib/router/handler-info.js
+++ b/lib/router/handler-info.js
@@ -154,9 +154,10 @@ function paramsMatch(a, b) {
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


sandstrom:

rwjblue:
I'm happy to land this if you can put together a test case that fails before this fix and is passing afterwards.

Due to the annoying rebase conflict issues, I'm going to close this for now and you can open a new PR when/if you would like to proceed.

PR #162, opened 2015-07-06
Correcting example function in README

r4j4h:
The code used to reference an undefined `posts` variable and did not use the passed parameter `object`. Later on down serialize is used again but does use its passed parameter, indicating that this one may be wrong.


stefanpenner:
good catch!

r4j4h:
Thanks :) and great work!

Issue #60, opened 2013-10-28
Multiple params in a single route path


KidkArolis:
I've noticed this doesn't work:

```
router.generate("foo", 1, 2);
```

if both of those params belong to the same route.

```
this.route("foo", {path: "/:param1/:param2"});
```

due to how only a single name is passed to `getMatchPointObject` (https://github.com/tildeio/router.js/blob/master/lib/router.js#L480).

Is this by design (e.g. related to models), or is this something that should/could be fixed?

jmeas:
Is this still an issue...?

stefanpenner:
@jmeas if you have time, adding a quick test to the test suite may both reveal if this issue is valid and provide another piece of coverage.

nathanhammond:
The tests from #164 proved that this is not an issue. Closing, will still attempt to land the PR.

PR #164, opened 2015-08-10
Add tests demonstrating that #60 is resolved.

jmeas:
This adds a test demonstrating that #60 is no longer an issue (if I read the issue right that is).

Lmk if you think that these tests belong somewhere else in test directory...they probably do :)

### DIFF
--- a/test/tests/router_test.js
+++ b/test/tests/router_test.js
@@ -18,6 +18,7 @@ module("The router", {
       });
       match("/posts", function(match) {
         match("/:id").to("showPost");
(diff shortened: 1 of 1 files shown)


nathanhammond:
These tests having never been merged are now badly out of date. We know this not to be an issue at this time. @jmeas would you like to rebase and land? If not I'll simply close this PR.

pixelhandler:
@jmeas @nathanhammond Here is a PR for the same, added `asserts` as needed, see https://github.com/tildeio/router.js/pull/200

PR #165, opened 2015-08-10
Add example history implementations to README.

### LINKED ISSUE #110: Examples of URL libs
From the README:

> Many other libraries do a good job of abstracting hash and pushState and working around known bugs in browsers.

Can you provide examples in the README of what some of these libraries are?

jmeas:
Resolves #110


rwjblue:
Thanks!

PR #166, opened 2015-08-13
Added isInitialSequence method

cacuario:
Since this seems like a frequent check, I added a method for this.


machty:
Sorry for the long delay, this seems ok but would you mind adding a test?

nathanhammond:
We've made it this long without adding this as new API surface area. I'd like to keep the API small as we continue to investigate how this library will change in the future. We very much appreciate your PR.

PR #167, opened 2015-08-17
Prepare for npm redux

kruppel:
This branches from @pzuraq's work in #149 .

The additional commit uses `broccoli-es6modules` for ES6 module transpilation and exports a UMD dist file instead of the collection of commonjs dist files.

Apologies in advance for opening the third PR for this request. I'd personally just like to see this go through and this seemed like the most reasonable way to advance the discussion. If other means are preferred, by all - err - means, let me know.


pzuraq:
:+1: looks solid to me, thanks for updating that @kruppel!

nathanhammond:

PR #168, opened 2015-08-20
Remove duplicate decode function

crsici:
I've found the issue when parameter value is '%' so my URL will be ...?parameterName=%25

At the first time at route-recognizer.js:482 
        path = decodeURI(path);
will decode the path to ...?parameterName=%
At the second attempt to decode by decodeURIComponent for each parameter pair, an exception will be thrown.

Uncaught URIError: URI malformed
    at decodeURIComponent (native)
    at Object.eval (eval at evaluate (unknown source), <anonymous>:1:1)
    at Object.InjectedScript._evaluateOn (<anonymous>:905:55)
    at Object.InjectedScript._evaluateAndWrap (<anonymous>:838:34)
    at Object.InjectedScript.evaluateOnCallFrame (<anonymous>:964:21)

Because it tries to decodeURIComponent('%')

### DIFF
--- a/vendor/deps/route-recognizer.js
+++ b/vendor/deps/route-recognizer.js
@@ -448,7 +448,7 @@ define("route-recognizer",
         var pairs = queryString.split("&"), queryParams = {};
         for(var i=0; i < pairs.length; i++) {
           var pair      = pairs[i].split('='),
-              key       = decodeURIComponent(pair[0]),
+              key       = pair[0],
               keyLength = key.length,
               isArray = false,
               value;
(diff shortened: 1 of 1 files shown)


krisselden:
Route recognizer should not be decoding the query string before it has been parsed.  Decoding is supposed to be done after all structure has been parsed, route recognizer should only decode parts of the url it does recognition on, basically the path segments after split('/'), it should be using unescape on these parts.  https://github.com/tildeio/router.js/issues/178

krisselden:
This issue should be opened on route-recognizer and fixed there.

PR #171, opened 2015-09-30
[WIP] Make model hooks run in parallel

nickiaconis:
Opening for discussion on how to support running model hooks in parallel. From an IRL discussion with @wycats and @nathanhammond, this should be built as an addon, so that apps must explicitly opt into the functionality and related consequences. I am laying out the required changes directly in the source so we can plan how to hook into the necessary parts of Ember from that addon.

Most of the changes to make this work happen here, but parallel model hooks require changes to ember.js as well: emberjs/ember.js#12415


nathanhammond:
I'm closing in favor of https://github.com/nickiaconis/ember-prefetch ... does that seem reasonable? I don't believe that any of these patches are still required once we have the router service stuff landed.

nickiaconis:
Agree. Thanks for closing. 🙂

PR #175, opened 2015-12-24
Include package dependencies; bower support

rmehlinger:
Adds RSVP, route-recognizer to package.json.
Creates a simple bower.json that can be used
for registration and installation, but does not
include devDependencies, as Broccoli is not on
Bower.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /.bundle
 /tmp
 /node_modules
+/bower_components
 
 dist/tests
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,27 @@
+{
+  "name": "router.js",
+  "description": "A lightweight JavaScript library is built on top of route-recognizer and rsvp.js to provide an API for handling routes",
+  "main": "router.js",
+  "authors": [
+    "Tilde",
+    "Inc."
+  ],
+  "license": "MIT",
+  "keywords": [
+    "router",
+    "route-recognizer",
+    "rsvp"
+  ],
+  "homepage": "https://github.com/tildeio/router.js",
+  "moduleType": [],
+  "ignore": [
+    "**/.*",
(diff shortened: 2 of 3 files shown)


mmun:
The NPM dependencies are good but we're trying to move away from Bower. Do you have a use case in mind?

rmehlinger:
Briefly, I work on a number of projects that use Bower and also include routerjs. Currently we have to do that by git hash, which is very inconvenient. I'd like to be able to `bower install --save router.js` so that I don't have to include NPM for what are otherwise bower-only projects.

jmeas:
im +1 to discourage use of Bower, but I'm also +1 to include the bower.json. lots of people still use it, and I think it's the responsibility of package maintainers to make it as easy as possible for many people to consume the package.

mmun:
I don't know what you mean by "otherwise include NPM". If you have bower installed then you also have NPM installed. What difference does it make to add a package.json and pull your dependencies out of node_modules rather than bower_components?

rmehlinger:
Yes, you have NPM installed (probably globally) if you have bower installed, but there are definitely projects I'm on where the preference is to manage frontend requirements through Bower. I understand that Bower seems to be on its way out as a package manager, but lots of projects still use it for their primary package manager and the amount of effort necessary to support it is minimal.

I'll also note that Ember itself includes a bower.json, which references this project via github URL rather than semver: https://github.com/emberjs/ember.js/blob/master/bower.json. 

This strikes me as less than ideal. RSVP and route-recognizer also both include a bower.json. So I guess I just don't see why routerjs shouldn't too.

nathanhammond:

We don't intend to surface this via publishing to Bower and have just removed Ember's Bower dependency on router.js. I recommend adopting npm for this.

Thank you so much for the PR! We appreciate your desire to help.

PR #180, opened 2016-03-30
Make Transition#urlMethod public

bendemboski:
Document this property and make it public so route hooks can detect the type (URL method) of transition that is occurring.

#169 documented the Transition class and specified several methods and properties as public or private. This is a follow-up in which I am proposing that we document the ```Transition#urlMethod``` property and make it public. The [Transition#method](https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L238) method that updates it is already public, so it seems to make sense to make it public as well, so there is a public way of reading it as well as writing it.

## Background

[This](https://github.com/emberjs/ember.js/issues/12824) issue describes a problem where a route that performs a redirect needs to know whether the transition's target URL is already in the history or not so it knows whether to call ```replaceWith()``` or ```transitionTo()``` in order to keep the navigation history correct.

The [workaround](https://github.com/emberjs/ember.js/issues/12824#issuecomment-172147220) described in the comments involves using the ```Transition#sequence``` property, which is currently private, only used for logging, and set based on an incrementing static class property that is not reset when the app resets (e.g. in unit tests). In short, a hack.

Making ```Transition#urlMethod``` public would provide a much cleaner workaround for the bug. Even ignoring that issue, it seems plausible that certain routes might want to make different decisions based on whether the transition they are handling is doing a replace, update, or not modifying the history.

### DIFF
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -76,7 +76,6 @@ Transition.currentSequence = 0;
 
 Transition.prototype = {
   targetName: null,
-  urlMethod: 'update',
   intent: null,
   params: null,
   pivotHandler: null,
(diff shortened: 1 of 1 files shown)


alexspeller:
Fix for the root cause of wanting this at https://github.com/tildeio/router.js/pull/197

bendemboski:
Awesome, thanks!

PR #181, opened 2016-04-17
Fix shaping

chadhietala:
Hmmm not really sure why the comments changed. Anyways, looks like there was much shape-shifting in this dude. 
image

@krisselden or @stefanpenner can you review.


dist/commonjs/router/router.js
```diff
@@ -55,7 +63,7 @@ function getTransitionByIntent(intent, isIntermediate) {
     }
 
     // No-op. No need to create a new transition.
-    return new Transition(this);
+    return this.activeTransition || new Transition(this);
```
stefanpenner: This change doesn't appear to be shaping related.

chadhietala: Yea I don't know what is going on here. Was not intended.

stefanpenner:
Ah. Then maybe some of the last changes forget to publish dist

lib/router/router.js
```diff
@@ -19,6 +19,13 @@ function Router(_options) {
   this.delegate = options.delegate || this.delegate;
   this.triggerEvent = options.triggerEvent || this.triggerEvent;
   this.log = options.log || this.log;
+  this.dslCallBacks = []; // NOTE: set by Ember
```
jmeas: hmmm...Ember-specific code in route-recognizer? 😕

jmeas:
What's the reasoning for these updatez?

dist/commonjs/router/router.js
jmeas: You may already know this, but with some git magic you could fix this. If you...

1. checked out a branch from master
2. built and PR'd that
3. came back here, reset the changes to dist
3. rebased this PR off that other PR

...it would make few a cleaner diff here and more atomic commit history.

chadhietala:
@jmeas It's important that you do not mutate the shape of the object post creation time. For each property on the object a hidden class is allocated for it. You should allocate these upfront so that the engine can optimize. The `Transition` class is probably the best example in this PR. If you create a `Transition` class and you only assign a portion of the fields you will end up with a hidden class layout table that looks like:

name | offset 
---------|---------
promise | 0
error | 1
params | 2

Now if you create a new `Transition` class that has one more field we now you have to allocate a new hidden class as the layout table looks like the following.

name | offset 
---------|---------
promise | 0
error | 1
params | 2
targetName | 3

In comparison if you "slot" all of the fields up front you end up using the same hidden class for all instances instead of having an explosion of types. Below is an illustration of this, also you might find these [articles](http://thibaultlaurens.github.io/javascript/2013/04/29/how-the-v8-engine-works/) [interesting](http://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html).


image

dist/commonjs/router/router.js
stefanpenner: @jmeas sounds reasonable, that way the diffs are cleaner when we have to do a bisect.
Unrelated to this PR, we should likely explore not committing built assets to the repo.

lib/router/router.js
stefanpenner: this sounds like a short-term thing, ideally ember should track this state itself in ember-router instance, a sibling to router instance.

jmeas:
Ah, okay, @chadhietala . I know a bit about V8 optimizations in regards to shaping – thanks for the detailed reply! :v:

lib/router/router.js
jmeas: Cool, cool.

rwjblue:
@chadhietala - OK, I just updated dist, can you rebase and remove the changes to `dist/` and only update the source files?

rwjblue:
Ping @chadhietala

chadhietala:
Tis done.

rwjblue:
Thanks @chadhietala!

PR #182, opened 2016-04-24
Introduce getSerializer option

trentmwillis:
Allows the logic for linking to a route (e.g., constructing urls) to be separated from the logic needed to enter a route.

This is a corollary to the [Route Serializers RFC from Ember.js](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md). The main goal here is to enable URLs/links for a route to be constructed _before_ a route's handler has been loaded. This will help enable the ability to lazy-load routes. The next step will be to allow `getHandler` to return a Promise value for a handler.

cc @dgeb @rwjblue

### DIFF
--- a/lib/router/handler-info/unresolved-handler-info-by-object.js
+++ b/lib/router/handler-info/unresolved-handler-info-by-object.js
@@ -25,7 +25,8 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
   serialize: function(_model) {
     var model = _model || this.context,
         names = this.names,
-        handler = this.handler;
+        handler = this.handler,
+        serializer = this.serializer || (handler && handler.serialize);
 
     var object = {};
     if (isParam(model)) {
@@ -34,8 +35,8 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
     }
 
     // Use custom serialize if it exists.
-    if (handler.serialize) {
-      return handler.serialize(model, names);
+    if (serializer) {
+      return serializer(model, names);
     }
 
     if (names.length !== 1) { return; }
--- a/lib/router/router.js
+++ b/lib/router/router.js
@@ -12,6 +12,7 @@ var pop = Array.prototype.pop;
 function Router(_options) {
   var options = _options || {};
   this.getHandler = options.getHandler || this.getHandler;
+  this.getSerializer = options.getSerializer || this.getSerializer;
   this.updateURL = options.updateURL || this.updateURL;
   this.replaceURL = options.replaceURL || this.replaceURL;
   this.didTransition = options.didTransition || this.didTransition;
@@ -29,7 +30,7 @@ function getTransitionByIntent(intent, isIntermediate) {
   var oldState = wasTransitioning ? this.activeTransition.state : this.state;
   var newTransition;
 
-  var newState = intent.applyToState(oldState, this.recognizer, this.getHandler, isIntermediate);
+  var newState = intent.applyToState(oldState, this.recognizer, this.getHandler, isIntermediate, this.getSerializer);
   var queryParamChangelist = getChangelist(oldState.queryParams, newState.queryParams);
 
   if (handlerInfosEqual(newState.handlerInfos, oldState.handlerInfos)) {
@@ -104,6 +105,8 @@ Router.prototype = {
   },
 
   getHandler: function() {},
+
+  getSerializer: function() {},
 
   queryParamsTransition: function(changelist, wasTransitioning, oldState, newState) {
     var router = this;
@@ -277,7 +280,7 @@ Router.prototype = {
     // Construct a TransitionIntent with the provided params
     // and apply it to the present state of the router.
     var intent = new NamedTransitionIntent({ name: handlerName, contexts: suppliedParams });
-    var state = intent.applyToState(this.state, this.recognizer, this.getHandler);
+    var state = intent.applyToState(this.state, this.recognizer, this.getHandler, null, this.getSerializer);
     var params = {};
 
     for (var i = 0, len = state.handlerInfos.length; i < len; ++i) {
@@ -297,7 +300,7 @@ Router.prototype = {
     });
 
     var state = this.activeTransition && this.activeTransition.state || this.state;
-    return intent.applyToState(state, this.recognizer, this.getHandler);
+    return intent.applyToState(state, this.recognizer, this.getHandler, null, this.getSerializer);
   },
 
   isActiveIntent: function(handlerName, contexts, queryParams, _state) {
@@ -330,7 +333,7 @@ Router.prototype = {
       contexts: contexts
     });
 
-    var newState = intent.applyToHandlers(testState, recogHandlers, this.getHandler, targetHandler, true, true);
+    var newState = intent.applyToHandlers(testState, recogHandlers, this.getHandler, targetHandler, true, true, this.getSerializer);
 
     var handlersEqual = handlerInfosEqual(newState.handlerInfos, testState.handlerInfos);
     if (!queryParams || !handlersEqual) {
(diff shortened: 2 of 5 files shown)


rwjblue:
LGTM

lib/router/handler-info/unresolved-handler-info-by-object.js
```diff
@@ -25,7 +25,8 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
   serialize: function(_model) {
     var model = _model || this.context,
         names = this.names,
-        handler = this.handler;
+        handler = this.handler,
+        serializer = this.serializer || (handler && handler.serialize);
```
MiguelMadero: This part seems a bit odd. If the idea of moving the serializer to the route is to allow lazy-loading, then the handler will only be available if the asset containing it has been loaded. Unless `getSerializer` always returns a function with default behavior or we use a flag to toggle between the old and the new behavior, this could lead to a different serialization logic used when an asset is loaded and when it's not. Considering that moving away from Route#serialize is relatively trivial, I suggest that both modes aren't supported for a given instance of a Router.

rwjblue: In [RFC#120](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md) we agreed to move serializers out of the route itself (and into the `Router.map`).  

> If the idea of moving the serializer to the route is to allow lazy-loading, then the handler will only be available if the asset containing it has been loaded.

This PR is *removing* the usage of `Route.prototype.serialize` if the serializer is already present (from the `Router.map` which is always available synchronously).

> Considering that moving away from Route#serialize is relatively trivial, I suggest that both modes aren't supported for a given instance of a Router.

Large apps take time to migrate, we *definitely* should not force them to migrate every usage at once. Gradual migrations are definitely better.

MiguelMadero: > This PR is removing the usage of Route.prototype.serialize if the serializer is already present (from the Router.map which is always available synchronously).

Maybe I'm reading it wrong, but let me try to explain the problem I have in mind. When the serializer isn't defined in Router.map, the route instance (handler) defines `serialize` and we lazy-load it. Before the route (handler) is loaded, we will fallback to the default serializer behavior, when the route is loaded, we will use `handler.serialize` instead.

MiguelMadero: Just moving explaining it in code. 

```
Router.map(function () {
  this.route('lazy-loaded-route', {path: '/lazy-loaded-route/:post_id'});
})

// path/routes/lazy-loaded-route
export default Ember.Route.extend({
  serialize: function () {
   return { post_id: 42 };
  }
});

{{link-to 'lazy-loaded-route' model}}
```

If the route is loaded, the link will be `lazy-loaded-route/42`, if the route is not loaded, the link will be `lazy-loaded-route/{model.id}`

trentmwillis: @MiguelMadero the scenario you are envisioning is correct. In order to properly support lazy-loading it will be a pre-requisite that developers use the new approach. It is something that we will likely add to the guides for using Engines and lazy-loading.

However, we can not simply drop the old approach as this would break existing applications and not follow semver.

MiguelMadero: > we can not simply drop the old approach as this would break existing applications and not follow semver.

I agree with that. What about a flag to remove the old behavior? I think that if we lazy-load we need to drop the old behavior to avoid the inconsistent behavior. Also, serialize is so rarely used that even for large apps we should be fine moving all at once.

rwjblue: If you use the older Route#serialize methods, you will already be getting a deprecation. Why do we need more layers?

MiguelMadero: I think that the combination of lazy-loading with Route#serialize leads to inconsistent behavior that can lead to hard to find bugs, that's why I think it would be better not to support this combination by allowing for a way to disable the use of Route#serialize.

nathanhammond: I'm in agreement with @MiguelMadero here. Any route which gets lazy-loaded should have its `serialize` hook replaced with an assertion, or fail to build. I'm pretty sure that we shouldn't handle that as part of router.js, though.

nathanhammond:
LGTM, in favor of shipping.

rwjblue:
Thanks for reviewing @nathanhammond, I agree.  Lets land and get this included for when we enable the `ember-route-serializers` feature (should be 2.7.0-beta.1), but before we have lazy loading in Engines.  This gives us a nice cycle of having the deprecation enabled before we even have the possibility of any danger...

lib/router/handler-info/unresolved-handler-info-by-object.js
```diff
@@ -34,8 +35,8 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
     }
 
     // Use custom serialize if it exists.
-    if (handler.serialize) {
-      return handler.serialize(model, names);
+    if (serializer) {
+      return serializer(model, names);
```
bendemboski: This introduced a breaking change. Any route that implemented the ```serialize``` hook and referenced ```this``` from within the hook is now broken, since the serialize method is called unbound. Maybe too late to do anything about it, and I guess that pattern is on its way out anyway, but thought I'd mention it so folks are aware.

nathanhammond: You're correct that this is a change in semantics. (Obviously. And you don't need me to agree with you to know that. 😄)

It's possible to "fix" in that we could [save off context here](https://github.com/tildeio/router.js/pull/182/files/d12ae9e58062fc7effe1557e13e704b3d58bf838#diff-fda34cdb7e594e6019ca204f39a9d00aR29) and then use `.call`. However with the route-serializers RFC that may be a bit surprising to people.

I'm personally of the opinion that I want serializers to be pure transforms, not based upon instance state. Given that, I would argue that having access to `this` was a bug itself, not a feature. Still, can you describe your use case so we can make a more informed decision?

(Also, congratulations! You may be the only person using this pattern in the entire Ember community considering the lag between release and report.)

bendemboski: Mine is not a use case that I'd use to argue for "fixing" this bug. For complicated reasons (that I would have probably avoided had I been as versed in Ember as I am now a year a half ago when I wrote it), my afterModel hook pushes some state into a service, and my serialize hook needs to recover it, so it does so via an injection on the route object. I suppose I could ```getOwner(model).lookup('service:...')``` rather than use an injection, but I think the whole setup that I have is kind of misguided. Anyway, I solved it by

```javascript
init() {
  this._super(...arguments);
  this.serialize = (model) => {
    // ...
  };
}
```

I agree with making the serializer a pure transform, because that would drive cleaner patterns rather than the mess I've gotten myself into. However, that being the case I think it's very counter-intuitive to have the serializer as a property of the router, because then it looks like an instance method hook like the other model hooks, and my expectation is that it is called with ```this``` bound to the route. But I guess you guys are already talking about deprecating it in favor of putting the serializer in the route map? Makes sense to me.

But now while it's still available as what looks like a route hook, I don't think it really makes sense to have it unbound. But, as you point out, I have a workaround and since that likely represents 100% of the people encountering this problem, I can see it making sense to just leave it since this form of serialize is going away anyway.

nathanhammond: I _really_ appreciate the thought you've put into this. We'll discuss this more and come to a conclusion. In fact you can follow along in [#dev-router](https://embercommunity.slack.com/messages/dev-router/) on the [Ember Community Slack](https://ember-community-slackin.herokuapp.com/) if you wish. 😄

PR #183, opened 2016-05-05
Introduce async support for getHandler

trentmwillis:
## READY FOR REVIEW

In order to support lazy-loading, router.js should not assume handlers returned from `getHandler` will be present synchronously. This PR introduces support for loading handlers asynchronously by returning a Promise from the `getHandler` function.

A common response to this idea is **why not load handlers upfront?** The reason to avoid loading handlers upfront is that they have the potential of being stateful and dependent on other constructs. This introduces a dependency tree that also must be loaded upfront with the handler, thus potentially negating the benefit of lazy loading.

This implementation reuses all existing tests that defined a `getHandler` method and has them passing while returning a Promise-based handler instead. One callout here is that `intermediateTransitionTo` passes under the assumption that it is used to transition into a synchronously loaded route.

TODO:
- [x] Rebase ontop of #182 
- [x] Simplify and make changes backwards-compatible
- [x] Audit tests to make sure everything is covered
- [x] Add documentation and clean up where appropriate
- [x] Squash commits


rwjblue:
It would be great if @machty had a chance to review (though not a blocker per-se).

lib/router/handler-info.js
```diff
@@ -36,22 +51,34 @@ HandlerInfo.prototype = {
...
-           .then(checkForAbort, null, this.promiseLabel("Check if aborted in 'model' hook"))
-           .then(afterModel, null, this.promiseLabel("After model"))
-           .then(checkForAbort, null, this.promiseLabel("Check if aborted in 'afterModel' hook"))
-           .then(becomeResolved, null, this.promiseLabel("Become resolved"));
+        becomeResolved = bind(this, this.becomeResolved,     payload),
+        self = this;
+
+    return Promise.resolve(this.handlerPromise, this.promiseLabel("Start handler"))
+            .then(function(handler) {
+              return Promise.resolve(handler)
```
rwjblue: Is this nesting needed? I could be missing something, but I would think this would work:

```js
return Promise.resolve(this.handlerPromise, this.promiseLabel("Start handler"))
    .then(checkForAbort, null, self.promiseLabel("Check for abort"))
    .then(beforeModel, null, self.promiseLabel("Before model"))
    .then(checkForAbort, null, self.promiseLabel("Check if aborted during 'beforeModel' hook"))
    .then(model, null, self.promiseLabel("Model"))
    .then(checkForAbort, null, self.promiseLabel("Check if aborted in 'model' hook"))
    .then(afterModel, null, self.promiseLabel("After model"))
    .then(checkForAbort, null, self.promiseLabel("Check if aborted in 'afterModel' hook"))
    .then(becomeResolved, null, self.promiseLabel("Become resolved"));
```

trentmwillis: So we can do that, the one issue with that approach (and the reason the code is this way) is if we have an error early on in the chain (for example [when checking `inaccessibleByURL`](https://github.com/tildeio/router.js/pull/183/files#diff-5863acacff50c33c3f6077ebdf56d4fbR53)), then we have to propagate it through the whole chain which is annoying.

lib/router/utils.js
```diff
@@ -11,6 +11,13 @@ if (!Array.isArray) {
 
 export var isArray = _isArray;
 
+/**
+  Determines if an object is Promise by checking if it is "thenable".
+**/
+export function isPromise(obj) {
+  return !!obj && (typeof obj === 'object' || typeof obj === 'function') && typeof obj.then === 'function';
```
rwjblue: Can you tweak to:

```js
return ((typeof obj === 'object' && obj !== null) || typeof obj === 'function') && typeof obj.then === 'function';
```

We would prefer to only do the comparison to `null` (instead of `!!obj`) when it is an `object`.

test/tests/query_params_test.js
```diff
@@ -3,8 +3,24 @@ import Router from "router";
...
+  {
+    name: 'Sync Get Handler',
+    getHandler: function(name) {
+      return handlers[name] || (handlers[name] = {});
+    }
+  },
+  {
+    name: 'Async Get Handler',
+    getHandler: function(name) {
+      return Promise.resolve(handlers[name] || (handlers[name] = {}));
```
rwjblue: To ensure we are async (and not just chaining on a `.then`able synchronously) can you change to:

```js
return new Promise(function(resolve) {
  setTimeout(function() {
    resolve(handlers[name] || (handlers[name] = {}));
  }, 1);
});
```

test/tests/transition_intent_test.js
```diff
@@ -7,11 +7,46 @@ import TransitionState from 'router/transition-state';
...
+scenarios.forEach(function(scenario) {
+
+// Asserts that a handler from a handlerInfo equals an expected valued.
+// Returns a promise during async scenarios to wait until the handler is ready.
+function assertHandlerEquals(handlerInfo, expected) {
+  if (!scenario.async) {
+    return equal(handlerInfo.handler, expected);
+  } else {
+    equal(handlerInfo.handler, undefined);
+    handlerInfo.handlerPromise.then(function(handler) {
```
rwjblue: I believe this should be returned?

test/tests/transition_intent_test.js
```diff
@@ -180,17 +203,16 @@ test("URLTransitionIntent applied to an already-resolved handlerInfo of differen
...
   var intent = new URLTransitionIntent({ url: '/foo/bar', });
-  var newState = intent.applyToState(state, recognizer, getHandler);
+  var newState = intent.applyToState(state, recognizer, scenario.getHandler);
   var handlerInfos = newState.handlerInfos;
 
   equal(handlerInfos.length, 2);
   ok(handlerInfos[0] !== startingHandlerInfo, "The starting foo resolved handlerInfo gets overridden because the new one has a different name");
   ok(handlerInfos[1] instanceof UnresolvedHandlerInfoByParam, "generated state consists of UnresolvedHandlerInfoByParam, 2");
-  equal(handlerInfos[1].handler, handlers.bar);
+  assertHandlerEquals(handlerInfos[1], handlers.bar);
```
rwjblue: Needs a `return`

rwjblue:
Just a few small nit-pick things, but this LGTM.  Can you update those things and squash?

test/tests/query_params_test.js
trentmwillis: I'm fairly certain Promise resolution has to be async but I will update for further guarantees (and would make it easy to test longer pauses).

rwjblue: I agree that I am just being paranoid, and your original code is certainly correct. Thank you for humoring me though.

trentmwillis: So turns out the tests setup a custom async handler for RSVP ([see here](https://github.com/tildeio/router.js/blob/master/test/tests/test_helpers.js#L10)), which seems to cause problems when used in conjunction with `setTimeout` since Backburner gets flushed independent of when the timeouts resolve.

I tend to think that we should proceed as is.

rwjblue: We chatted about this in slack, and decided to leave these looped tests as is and add a single new async only test that uses `setTimeout` and confirms a transition into a lazily resolved route works properly.

Future work will be to refactor these tests to not assume that all async is pseudo/fake async in the BB queue.

lib/router/utils.js
```diff
@@ -11,6 +11,13 @@ if (!Array.isArray) {
 
 export var isArray = _isArray;
 
+/**
+  Determines if an object is Promise by checking if it is "thenable".
+**/
+export function isPromise(obj) {
+  return ((typeof obj === 'object' && typeof obj !== null) || typeof obj === 'function') && typeof obj.then === 'function';
```
rwjblue: Should just be `obj !== null`, not `typeof obj !== null`

test/tests/async_get_handler_test.js
```diff
@@ -0,0 +1,54 @@
+import Router from 'router';
+import { Promise } from "rsvp";
+
+QUnit.module('Async Get Handler', {
```
rwjblue: Can you add a comment here to explain that we are intentionally using `QUnit.module` to avoid the async shenanigans in the `module` that is imported in the other tests?

PR #184, opened 2016-07-19
Bump QUnit to last 1.x release

trentmwillis:
Step 1 on the road to a healthier test suite.

All the test changes here have to do with the fact that tests were behaving as though they were async when in reality they weren't. With the QUnit improvements, this caused errors because QUnit would attempt to restart and complain that it never stopped.

(Note: we should probably not check QUnit into the repo, but I'll leave changing the build pipeline for another time.)


rwjblue:
Thanks @trentmwillis!

PR #185, opened 2016-07-20
Improve linting

trentmwillis:
Step 2 in improving test health of the repo.

I noticed during #183 that there are a lot of unused variables in the codebase so I enabled that check in JSHint and fixed all the issues. The second commit removes `jshint` and `grunt-contrib-jshint` as dependencies since they are redundant with `broccoli-jshint`.


stefanpenner:
@trentmwillis awesome, thank you.

PR #186, opened 2016-07-20
Restrict globals usage and remove QUnit globals usage

trentmwillis:
Step 3 in improving test help.

The number of allowed globals was too high. In particular, the tests all used the QUnit globals which have been deprecated and removed in 2.x. This helps keep tests more atomic.

Remaining piece after this will be to extract the pseudo-async backburner logic to give better guarantees about testing actually async behavior.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,24 +1,9 @@
 {
     "predef": [
-        "console",
-        "define",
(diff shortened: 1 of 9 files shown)


rwjblue:
Thank you!

PR #187, opened 2016-08-02
Get handlers lazily

trentmwillis:
Currently, `getHandler` is called before constructing a `HandlerInfo` object. This
results in work being done upfront even if the `handler` property of the
`HandlerInfo` is never accessed. This change invokes `getHandler` only when the
value is actually needed.

This is particularly needed in the case of `getHandler` fetching lazily-loaded
routes. Previously lazy routes would be fetched even if the router was just
generating a URL; a task which can be done without the whole handler.

Note: This change relies on getters/setters which implies IE9+ support (which is in line with Ember's support policy, but I am unsure about this library itself).

### DIFF
--- a/lib/router/handler-info.js
+++ b/lib/router/handler-info.js
@@ -1,21 +1,29 @@
 import { bind, merge, promiseLabel, applyHook, isPromise } from './utils';
 import Promise from 'rsvp/promise';
 
+var DEFAULT_HANDLER = Object.freeze({});
+
 function HandlerInfo(_props) {
   var props = _props || {};
-  var name = props.name;
 
-  // Setup a handlerPromise so that we can wait for asynchronously loaded handlers
-  this.handlerPromise = Promise.resolve(props.handler);
+  // Set a default handler to ensure consistent object shape
+  this._handler = DEFAULT_HANDLER;
 
-  // Wait until the 'handler' property has been updated when chaining to a handler
-  // that is a promise
-  if (isPromise(props.handler)) {
-    this.handlerPromise = this.handlerPromise.then(bind(this, this.updateHandler));
-    props.handler = undefined;
-  } else if (props.handler) {
-    // Store the name of the handler on the handler for easy checks later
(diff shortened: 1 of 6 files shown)


trentmwillis:
Unsure why tests are failing. Everything passes in browser and builds fine. Investigating...

lib/router/handler-info.js
```diff
@@ -24,7 +27,52 @@ function HandlerInfo(_props) {
 
 HandlerInfo.prototype = {
   name: null,
-  handler: null,
+
+  getHandler: function() {},
+
+  get handler() {
+    if (this.hasOwnProperty('_handler')) {
```
rwjblue: Why do we need the has own check here? Is there a scenario where we have inherited a `_handler`?

Would be easier/better?

```js
if (this._handler) { return this._handler; }
```

Also, `_handler` needs to be initialized to `undefined` or `null` in the constructor to avoid shaping issues (which I think would make the `hasOwnProperty` check fail to work since it would have the property but it would be undefined).

trentmwillis: I used `hasOwnProperty` since it might be `undefined`, which makes checks tricky. Basically we want "this has been set, but the type of value is ambiguous".

trentmwillis:
Tests were failing due to an old version of Phantom (1.9.2), bumping `grunt-contrib-qunit` fixes the issue.

trentmwillis:
@rwjblue updated this to fix the shaping.

lib/router/handler-info.js
```diff
@@ -24,7 +29,60 @@ function HandlerInfo(_props) {
 
 HandlerInfo.prototype = {
   name: null,
-  handler: null,
+
+  getHandler: function() {},
+
+  _handler: DEFAULT_HANDLER,
```
rwjblue: I think this needs to be in the constructor itself.  We want to shape the instance not the prototype...

rwjblue:
This is good once that last nit-pick is addressed.

Issue #188, opened 2016-08-13
#187 seems to crash PhantomJS 1.9 with a syntax error


simonihmig:
I noticed with a PR on one of my addons (https://github.com/kaliber5/ember-bootstrap/pull/123) that builds in TravisCI with ember canary were suddenly broken. They were actually not failing (which would be acceptable), but they seem to crash the whole build, TravisCI says: 

> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
>
> The build has been terminated

When running `ember try:each`, the canary build shows this:

```
not ok 1 PhantomJS 1.9 - Global error: SyntaxError: Parse error at http://localhost:7357/assets/vendor.js, line 75982
    ---
        Log: |
            { type: 'error',
              text: 'SyntaxError: Parse error at http://localhost:7357/assets/vendor.js, line 75982\n' }
    ...
not ok 2 PhantomJS 1.9 - Global error: Error: Could not find module ember-metal/core required by: ember-testing/index at http://localhost:7357/assets/test-support.js, line 62
    ---
        Log: |
            { type: 'error',
              text: 'Error: Could not find module ember-metal/core required by: ember-testing/index at http://localhost:7357/assets/test-support.js, line 62\n' }
    ...
not ok 3 PhantomJS 1.9 - Global error: ReferenceError: Can't find variable: define at http://localhost:7357/assets/dummy.js, line 24
    ---
        Log: |
            { type: 'error',
              text: 'ReferenceError: Can\'t find variable: define at http://localhost:7357/assets/dummy.js, line 24\n' }
    ...
not ok 4 PhantomJS 1.9 - Global error: ReferenceError: Can't find variable: define at http://localhost:7357/assets/tests.js, line 24
    ---
        Log: |
            { type: 'error',
              text: 'ReferenceError: Can\'t find variable: define at http://localhost:7357/assets/tests.js, line 24\n' }
    ...
not ok 5 PhantomJS 1.9 - Global error: ReferenceError: Can't find variable: EmberENV at http://localhost:7357/1738/tests/index.html?hidepassed, line 39
    ---
        Log: |
            { type: 'error',
              text: 'ReferenceError: Can\'t find variable: EmberENV at http://localhost:7357/1738/tests/index.html?hidepassed, line 39\n' }
    ...
``` 
Here the Travis logs from two addons, the latter with no code change for several weeks:
* https://travis-ci.org/kaliber5/ember-bootstrap/builds/151867897
[...]

PR #190, opened 2016-08-15
Use Object.defineProperty over getter/setter also for handlerPromise

simonihmig:
@rwjblue I think your PR #189 missed the `handlerPromise` property. Was still getting a PhantomJS crash as reported in #188 , even with the lastest canary build. This should hopefully fix it! :)

### DIFF
--- a/lib/router/handler-info.js
+++ b/lib/router/handler-info.js
@@ -56,20 +56,6 @@ HandlerInfo.prototype = {
 
   _handlerPromise: undefined,
 
-  get handlerPromise() {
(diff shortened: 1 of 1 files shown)


rwjblue:
Dagnabbit, thanks for double checking this. Would you mind submitting an update PR to Ember also?

simonihmig:
Sure, will do so!

simonihmig:
For the record: just rerun my addons' builds to make sure the error reported in #188 is gone... and it is!

PR #191, opened 2016-08-15
Document `Transition#isExiting` as public

bantic:
It appears as though the `Transition#isExiting` is intended to be public. This PR marks it as public, adds documentation for it, and adds tests for the `isExiting` method with both string and handler-object arguments for transitions between sibling and nested routes. Previously there were only a few tests that passed handler objects to `isExiting` and none that passed handler names (strings).

### DIFF
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -95,6 +95,16 @@ Transition.prototype = {
 
   isTransition: true,
 
+  /**
+    Check whether this transition is exiting the given `handler`.
+    The `handler` argument can be the the handler object itself or the name
+    of the handler.
+
+    @method isExiting
+    @param {String|Object} handler The handler's name or object.
+    @return {Boolean}
+    @public
+  */
   isExiting: function(handler) {
     var handlerInfos = this.handlerInfos;
     for (var i = 0, len = handlerInfos.length; i < len; ++i) {
--- a/test/tests/router_test.js
+++ b/test/tests/router_test.js
@@ -660,6 +660,98 @@ test("Moving to a new top-level route triggers exit callbacks", function(assert)
   }, shouldNotHappen(assert));
 });
 
+test("Transition#isExiting is correct when moving between sibling routes", function(assert) {
+  assert.expect(9);
+
+  var index = {
+    setup: function(_, transition) {
+      assert.ok(!transition.isExiting(index), "setup index (handler)");
+      assert.ok(!transition.isExiting("index"), "setup index (string)");
+    },
+    exit: function(transition) {
+      assert.ok(transition.isExiting(index), "exit index (handler)");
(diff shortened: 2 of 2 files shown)


mixonic:
> It appears as though the Transition#isExiting is intended to be public.

@bantic can you clarify why you believe it is intended to be public?

bantic:
> can you clarify why you believe it is intended to be public?

Sure. The method seems to have no internal use, either in this repo or in Ember.js. The only usage of it in this repo is in [one test](https://github.com/tildeio/router.js/blob/b921fd3407064a77bd99f4f4eddc4377592ac93b/test/tests/router_test.js#L628-L634), and I don't see it used in Ember.js at all.

That's admittedly not the strongest evidence, but if it isn't meant to be used publicly it appears to have no other use and should be removed, deprecated and/or explicitly marked private.

bantic:
cc @machty your commit (from a little while ago) https://github.com/tildeio/router.js/commit/2449b5af36df64d2017b07f224ba414994b1ef25 explicitly exposed `isExiting`. Do you think it makes sense to be marked "public"? Is it relied upon by some other library?

machty:
I think we should probably just completely remove it. I think it's an artifact of a first stab at implementing `resetController` in Ember, which receives an `isExiting` bool, but router.js doesn't use `Transition#isExiting` to determine this value, but determines it here: https://github.com/tildeio/router.js/blob/8000486a3c10036e08f1ceb742bd4b0b4f928996/lib/router/router.js#L449

I highly doubt anyone who uses router.js ever used it and I definitely don't think there's any use keeping it around, so I vote remove.

mitchlloyd:
Oh, seemed like a useful API.

Is there an alternative pattern to accomplish something like this in an Ember Route?

```js
willTransition(transition) {
  if (transition.isExiting('thisRoutesName')) {
    transition.abort();
    alert('not so fast!');
  }
}
```

mitchlloyd:
I just noticed that `resetController` gets the `transition` as an argument so this would work. Seems a little odd semantically, but has everything I need there.

machty:
@mitchlloyd hmm maybe you're right. I guess we don't really know if people ended up finding that method and making use of it. I know for instance a lot of people started digging around the transition/state/handlerInfos object to find what they needed even though those were considered "private". It's just unfortunate that due to router.js's microlib nature it/I was able to sneak through these APIs that people might now rely on. 

So I'm not sure what to do. Maybe the safest thing is to leave it in and document it, but ideally it shouldn't have been created to begin with. Hopefully going forward any microlib dependencies of Ember will go through the same RFC process (or perhaps Ember should more aggressively hide raw microlib API objects behind Ember-specific objects).

mitchlloyd:
Giving it the ol' "intimate API" treatment and deprecating is another option.

wagenet:
Closing due to staleness.

Issue #193, opened 2016-09-22
Chain handlerPromise to parent


trentmwillis:
It seems like it could be beneficial to chain a HandlerInfo's `handlerPromise` to the `handlerPromise` of the HandlerInfo that comes before it (the "parent" if you will). This would help guarantee that any callbacks chained to `handlerPromise` are resolved from top-to-bottom.

I imagine this would be relatively straightforward to implement. One drawback, however, is that it now creates a link between HandlerInfos were previously they were disjoint.

See https://github.com/nickiaconis/ember-prefetch/issues/14 for one potential use case.

cc @nathanhammond @nickiaconis

rwjblue:
👍 - This seems like something we should fix to me. Specifically, it will be required to handle the "lazy engine within a lazy engine" case...

trentmwillis:
So turns out chaining the Promises together without eagerly calling `getHandler` (something we want to avoid for lazy loading) is very difficult. That said, because we resolve HandlerInfos one at a time during a transition, we are guaranteed that all the first class functionality will execute in order. I will add a test to verify this behavior.

Anyone wishing to use `handlerPromise` outside of this, will need to chain the Promises manually. This is better because we can assume that if you're accessing `handlerPromise` then you are fine with `getHandler` being called. Making that assumption internally is much more difficult.

trentmwillis:
Behavior verified by #195.

PR #195, opened 2016-10-18
Add test to verify execution order of hooks in async handlers

trentmwillis:
Verifies the behavior we want out of https://github.com/tildeio/router.js/issues/193.

cc/ @rwjblue @nathanhammond


rwjblue:
Thanks for getting this scenario under test @trentmwillis!

PR #197, opened 2016-10-28
Fix validation stage redirection behaviour.

alexspeller:
This has been broken forever AFAIK, and is surprising to a lot of people. In
fact, even the ember guides recommend [using `this.transitionTo` for redirecting](https://guides.emberjs.com/v2.9.0/routing/redirection/)

This is in fact broken. If you use `transitionTo` as a redirect strategy,
then you get an extra history entry if you hit the route as the first route in
your app. This breaks the back button, as the back button takes you to the route
with the redirect, then immediately redirects back to the page you're on.
Maybe you can go back if you hammer back button really quickly, but you have to
hammer it loads super quick. Not a good UX.

`replaceWith` works if you use it to redirect from a route that's your
initial transition. However if you use it to redirect and you hit that route
from some way _after_ the initial app load, then at the point that the
replaceWith is called, you are still on the same URL. For example, if you are on
`/` and you click a link to `/foo`, which in it's model hook redirects to `/bar`
using `replaceWith`, at the time replaceWith is called, your current url is `/`.

This means `/` entry gets removed from your history entirely. Clicking back
will take you back to whatever page you were on before `/`, which often isn't
even your app, maybe it's google or something. This breaks the back button
again.

This commit should do the correct thing in all cases, allowing replaceWith and
transitionTo outside of redirects as specified by the developer but only allowing
transitionTo or replaceWith in redirects in a way that doesn't break the back
button.

### DIFF
--- a/lib/router/router.js
+++ b/lib/router/router.js
@@ -26,6 +26,7 @@ function Router(_options) {
   this.oldState = undefined;
   this.currentHandlerInfos = undefined;
   this.state = undefined;
+  this.currentSequence = 0;
 
   this.recognizer = new RouteRecognizer();
   this.reset();
@@ -59,7 +60,7 @@ function getTransitionByIntent(intent, isIntermediate) {
   }
 
   // Create a new transition to the destination route.
-  newTransition = new Transition(this, intent, newState);
+  newTransition = new Transition(this, intent, newState, undefined, this.activeTransition);
 
   // Abort and usurp any previously active transition.
   if (this.activeTransition) {
@@ -625,7 +626,27 @@ function updateURL(transition, state/*, inputUrl*/) {
     params.queryParams = transition._visibleQueryParams || state.queryParams;
     var url = router.recognizer.generate(handlerName, params);
 
-    if (urlMethod === 'replace') {
+    // transitions during the initial transition must always use replaceURL.
+    // When the app boots, you are at a url, e.g. /foo. If some handler
+    // redirects to bar as part of the initial transition, you don't want to
+    // add a history entry for /foo. If you do, pressing back will immediately
+    // hit the redirect again and take you back to /bar, thus killing the back
+    // button
+    var initial = transition.isCausedByInitialTransition;
+
+    // say you are at / and you click a link to route /foo. In /foo's
+    // handler, the transition is aborted using replacewith('/bar').
+    // Because the current url is still /, the history entry for / is
+    // removed from the history. Clicking back will take you to the page
+    // you were on before /, which is often not even the app, thus killing
+    // the back button. That's why updateURL is always correct for an
+    // aborting transition that's not the initial transition
+    var replaceAndNotAborting = (
+      urlMethod === 'replace' &&
+      !transition.isCausedByAbortingTransition
+    );
+
+    if (initial || replaceAndNotAborting) {
       router.replaceURL(url);
     } else {
       router.updateURL(url);
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -16,7 +16,7 @@ import { trigger, slice, log, promiseLabel } from './utils';
   @param {Object} error
   @private
  */
-function Transition(router, intent, state, error) {
+function Transition(router, intent, state, error, previousTransition) {
   var transition = this;
   this.state = state || router.state;
   this.intent = intent;
@@ -40,6 +40,18 @@ function Transition(router, intent, state, error) {
     return;
   }
 
+  // if you're doing multiple redirects, need the new transition to know if it
+  // is actually part of the first transition or not. Any further redirects
+  // in the initial transition also need to know if they are part of the
+  // initial transition
+  this.isCausedByAbortingTransition = !!previousTransition;
+  this.isCausedByInitialTransition = (
+    previousTransition && (
+      previousTransition.isCausedByInitialTransition ||
+      previousTransition.sequence === 0
+    )
+  );
+
   if (state) {
     this.params = state.params;
     this.queryParams = state.queryParams;
@@ -58,16 +70,9 @@ function Transition(router, intent, state, error) {
       this.pivotHandler = handlerInfo.handler;
     }
 
-    this.sequence = Transition.currentSequence++;
-    this.promise = state.resolve(checkForAbort, this)['catch'](function(result) {
-      if (result.wasAborted || transition.isAborted) {
-        return Promise.reject(logAbort(transition));
-      } else {
-        transition.trigger('error', result.error, transition, result.handlerWithError);
-        transition.abort();
-        return Promise.reject(result.error);
-      }
-    }, promiseLabel('Handle Abort'));
+    this.sequence = router.currentSequence++;
+    this.promise = state.resolve(checkForAbort, this)['catch'](
+      catchHandlerForTransition(transition), promiseLabel('Handle Abort'));
   } else {
     this.promise = Promise.resolve(this.state);
     this.params = {};
@@ -80,7 +85,18 @@ function Transition(router, intent, state, error) {
   }
 }
 
-Transition.currentSequence = 0;
+function catchHandlerForTransition(transition) {
+  return function(result) {
+    if (result.wasAborted || transition.isAborted) {
+      return Promise.reject(logAbort(transition));
+    } else {
+      transition.trigger('error', result.error, transition, result.handlerWithError);
+      transition.abort();
+      return Promise.reject(result.error);
+    }
+  };
+}
+
 
 Transition.prototype = {
   targetName: null,
--- a/test/tests/router_test.js
+++ b/test/tests/router_test.js
@@ -3327,6 +3327,660 @@ test("intermediateTransitionTo() forces an immediate intermediate transition tha
   counterAt(7, "original transition promise resolves");
 });
 
+
+test("Calling transitionTo during initial transition in validation hook should use replaceURL", function(assert) {
+  assert.expect(4);
+  map(assert, function(match) {
+    match("/foo").to('foo');
+    match("/bar").to('bar');
+  });
+
+  var fooModelCount = 0, barModelCount = 0;
+
+  router.updateURL = function(updateUrl) {
+    url = updateUrl;
+    assert.ok(false, "The url was not correctly replaced on initial transition");
+  };
+
+  router.replaceURL = function(replaceURL) {
+    url = replaceURL;
+    assert.ok(true, "The url was replaced correctly on initial transition");
+  };
+
+  var fooHandler = {
+    model: function() {
+      fooModelCount++;
+      router.transitionTo('/bar');
+    }
+  };
+
+  var barHandler = {
+    model: function() {
+      barModelCount++;
+    }
+  };
+
+  handlers = {
+    foo: fooHandler,
+    bar: barHandler
+  };
+
+  transitionTo(router, '/foo');
+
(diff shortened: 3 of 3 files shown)


alexspeller:
@stefanpenner you can retire `ember-redirect-to` now :)

lib/router/router.js
```diff
@@ -63,6 +63,13 @@ function getTransitionByIntent(intent, isIntermediate) {
 
   // Abort and usurp any previously active transition.
   if (this.activeTransition) {
+    // if you're doing multiple redirects, need the new transition to know if it
+    // is actually part of the first transition or not. Any further redirects
+    // in the initial transition also need to know if they are part of the
+    // initial transition
+    newTransition.isCausedByInitialTransition = this.activeTransition.isCausedByInitialTransition|| this.activeTransition.sequence === 0;
+    newTransition.isCausedByAbortingTransition = true;
```
trentmwillis: Can we set these in the Transition constructor? At the very least, we should set them to `undefined` in the constructor to avoid mutating the shape of the transition object.

lib/router/router.js
```diff
@@ -63,6 +63,13 @@ function getTransitionByIntent(intent, isIntermediate) {
 
   // Abort and usurp any previously active transition.
   if (this.activeTransition) {
```
trentmwillis: Can we save off a reference to `this.activeTransition`? Would help keep some of the lines below shorter.

lib/router/router.js
```diff
@@ -63,6 +63,13 @@ function getTransitionByIntent(intent, isIntermediate) {
...
+    // if you're doing multiple redirects, need the new transition to know if it
+    // is actually part of the first transition or not. Any further redirects
+    // in the initial transition also need to know if they are part of the
+    // initial transition
+    newTransition.isCausedByInitialTransition = this.activeTransition.isCausedByInitialTransition|| this.activeTransition.sequence === 0;
```
trentmwillis: Nitpick: Space before the `||`

lib/router/router.js
```diff
@@ -625,10 +632,32 @@ function updateURL(transition, state/*, inputUrl*/) {
...
+      // transitions during the initial transition must always use replaceURL.
+      // When the app boots, you are at a url, e.g. /foo. If some handler
+      // redirects to bar as part of the initial transition, you don't want to
+      // add a history entry for /foo. If you do, pressing back will immediately
+      // hit the redirect again and take you back to /bar, thus killing the back
+      // button
       router.replaceURL(url);
     } else {
-      router.updateURL(url);
+      if (urlMethod === 'replace' && !aborting) {
```
trentmwillis: Unsure how others will feel about this, but I'd prefer doing:

```js
if (initial) {
} else if (urlMethod === 'replace' && !aborting) {
} else {
}
```
Instead of the nesting you currently have.

test/tests/router_test.js
```diff
@@ -3327,6 +3329,672 @@ test("intermediateTransitionTo() forces an immediate intermediate transition tha
   counterAt(7, "original transition promise resolves");
 });
 
+
+test("Calling transitionTo during initial transition in validation hook should use replaceURL", function(assert) {
+  Transition.currentSequence = 0;
```
trentmwillis: Tangential to this PR, we should probably store this on the Router instance if possible.

alexspeller:
@trentmwillis thanks for detailed review, I made some improvements based on your comments, including moving currentSequence to router - that was bugging me too and it was a minor change

lib/router/router.js
```diff
@@ -625,7 +626,28 @@ function updateURL(transition, state/*, inputUrl*/) {
...
+    // removed from the history. Clicking back will take you to the page
+    // you were on before /, which is often not even the app, thus killing
+    // the back button. That's why updateURL is always correct for an
+    // aborting transition that's not the initial transition
+    var replaceAndNotAborting = (
+      urlMethod === 'replace' &&
+      !transition.isCausedByAbortingTransition
+    );
+
+
```
trentmwillis: Nitpick: Remove the extra line here

rwjblue:
@alexspeller - What happens to folks that are already working around these bugs? I _think_ that existing workarounds continue to work fine (but are just not needed), but I'd like to confirm.

alexspeller:
@rwjblue correct, this makes no difference to anyone using workarounds. This patch ensures that the right thing is always done. If the right thing is being done already there is no difference

rwjblue:
@nathanhammond - r?

lib/router/transition.js
```diff
@@ -16,7 +16,7 @@ import { trigger, slice, log, promiseLabel } from './utils';
   @param {Object} error
   @private
  */
-function Transition(router, intent, state, error) {
+function Transition(router, intent, state, error, previousTransition) {
```
nathanhammond: Must we pass the entire transition forward? It appears that we only need to identify if this is the initial transition. As it stands it seems like this might be closing over a lot of transitions which would keep them dormant in memory longer than we care for.

(Need somebody more familiar to confirm/deny, just a quick thought.)

test/tests/router_test.js
```diff
@@ -3327,6 +3327,660 @@ test("intermediateTransitionTo() forces an immediate intermediate transition tha
...
+
+  assert.equal(url, '/baz');
+  assert.equal(bazModelCount, 1, 'Baz model should be called once');
+  assert.equal(fooModelCount, 0, 'Foo model should not be called');
+  assert.equal(barModelCount, 0, 'Bar model should not be called');
+
+  transitionTo(router, '/foo');
+
+  assert.equal(url, '/baz');
+  assert.equal(bazModelCount, 2, 'Baz model should be called thrice');
```
nathanhammond: Something is wrong on this line, unsure whether it's the message or the expectation.

test/tests/router_test.js
```diff
@@ -3327,6 +3327,660 @@ test("intermediateTransitionTo() forces an immediate intermediate transition tha
...
+  var fooModelCount = 0, barModelCount = 0, bazModelCount = 0;
+
+  router.updateURL = function(updateUrl) {
+    url = updateUrl;
+    assert.ok(true, "updateURL should be used");
+  };
+
+  router.replaceURL = function(replaceURL) {
+    url = replaceURL;
+    assert.ok(false, "replaceURL should not be used");
```
nathanhammond: I love how you're partially following ye ol' @tomhuda style of "English? double quotes. Code? single quotes." :trollface: 

If you want to map all quote behavior in a specific way that might be worth doing. I didn't review the non-diff so I don't know how consistent it is currently.

test/tests/router_test.js
```diff
@@ -3327,6 +3327,660 @@ test("intermediateTransitionTo() forces an immediate intermediate transition tha
...
+  assert.equal(fooModelCount, 0, 'Foo model should not be called');
+  assert.equal(barModelCount, 0, 'Bar model should not be called');
+
+  transitionTo(router, '/foo');
+
+  assert.equal(replaceUrlCount, 0, 'replaceURL should not be used');
+  assert.equal(updateUrlCount, 2, 'updateURL should be used for subsequent transition');
+  assert.equal(url, '/baz');
+  assert.equal(bazModelCount, 2, 'Baz model should be called once');
+  assert.equal(fooModelCount, 1, 'Foo model should not be called');
```
nathanhammond: This one and the line below it need messaging adjustments.

lib/router/transition.js
rwjblue: I tend to agree with @alexspeller here, but might make sense for @stefanpenner / @krisselden  / other @tildeio/ember-core folks to sanity check also...

alexspeller:
Fixed test labelling, think I've addressed saving previous transition issue, not addressing quoting style currently

lib/router/transition.js
```diff
@@ -58,7 +70,7 @@ function Transition(router, intent, state, error) {
       this.pivotHandler = handlerInfo.handler;
     }
 
-    this.sequence = Transition.currentSequence++;
+    this.sequence = router.currentSequence++;
     this.promise = state.resolve(checkForAbort, this)['catch'](function(result) {
```
nathanhammond: Pretty sure this function catches the transition in the closure.

rwjblue: Ya, could be (stupid VM's doing smart things).  Without looking terribly deeply here, it seems that we might be able to avoid using closure scope stuff (and avoid generating a new function for each transition created too), by elevating this to a private prototype method (and moving `checkForAbort`) and calling it from the constructor..

nathanhammond: I'm still concerned about this accidentally trolling us in the future. One of us won't be paying attention and will merge a PR which accidentally again captures things in a closure. In general I'm hesitant to pass around "huge" (by reference) objects.

rwjblue: Hmm. We definitely need to figure out the flags to know which method to use, do you have suggestions that would allow us to accomplish both objectives?

nathanhammond: Actually right now I'm just being a jerk and identifying problems without coming up with proposed solutions. 😛 Addressing this one problem (for example, as you proposed) and then adding a comment to remind us to look out for this behavior is probably just fine.

I'd still like some GC/VM wizards to chime in on anything else we may have missed...

alexspeller: So to be clear, right now the best thing for me to do is to refactor that function in the transition constructor to a prototype method, so that it doesn't risk closing over the arguments passed in to the transition constructor?

FWIW I hadn't thought that it would close over variables that you don't explicitly use in the function body, however that's based on chrome's behaviour and not a deep understanding of the various js VMs ember supports

nathanhammond: That function is lexically scoped in a place where it can reach the previous transition which means it can't be cleaned up until that property is reset.

Yes, we need to refactor it such that the definition of this section happens outside of the constructor. This will have the bonus win of being more performant.

lib/router/transition.js
ef4: I agree that this doesn't affect retention.

krisselden: @alexspeller confirm, you aren't retaining it so it does not affect GC

krisselden: @alexspeller I spoke too soon, there is a closure to this context being created.

rwjblue: Thanks for reviewing / double checking @krisselden.

@nathanhammond - Are you happy to move forward with that confirmation, or should we do more tests (or refactor anyways)?

krisselden: on more detailed review, it looks like `transition` is the only property that would be added to the closure.  It is a little weird to do this in a condition.

I'd likely make a function that took `transition`as an arg that returned a error handler.  This limits the scope and shape of the closure context.

alexspeller:
@krisselden I believe [this is what you meant, avoiding the creation of a closure in the constructor now](https://github.com/tildeio/router.js/pull/197/commits/22a2afd69ee57d8e27df799f24ff86dc3edfeec6#diff-89feb243989c9fd30a082f8039300715R75)

@rwjblue @nathanhammond @trentmwillis AFAIK this should be ready for merge based on the feedback given above, let me know if there's anything else 😄

nathanhammond:
@alexspeller Thank you for keeping this thing moving with all of the deep JS internals review. I'm grateful for the detailed attention you paid to this and for getting this solved at the appropriate layer.

@krisselden I always learn things when you show up on threads, I appreciate your detailed analysis of what's going on in JS Engine land both in this thread and on Slack.

This change now appears to be rock solid, let land it! 🎉

PR #198, opened 2016-10-29
Fix queryParamsOnly being correct on transitions.

alexspeller:
This property has been set incorrectly on some transitions, notably when the
transition is caused by the Router#refresh() method in queryParamsDidChange.

This is the method ember uses to implement `refreshModel` option of query params,
so fixing this bug allows `refreshModel` to be used in conjunction with
`replace` option for query params, currently it is either-or.

Resolves emberjs/ember.js#10577 and emberjs/ember.js#11843

### DIFF
--- a/lib/router/router.js
+++ b/lib/router/router.js
@@ -45,6 +45,7 @@ function getTransitionByIntent(intent, isIntermediate) {
     if (queryParamChangelist) {
       newTransition = this.queryParamsTransition(queryParamChangelist, wasTransitioning, oldState, newState);
       if (newTransition) {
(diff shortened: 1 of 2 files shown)


alexspeller:
@mike-north this is relevant to your interests.

rwjblue:
This looks good to me, thanks for tackling it!

I'll leave it open for a day or so and give others a chance to review too...

nathanhammond:
@rwjblue I'm still 👍 on this.

darylalexjohnson:
@nathanhammond  good news! Thanks

engwan:
+1 Any update on this?

PR #199, opened 2016-11-03
Improve nesting documentation

### LINKED ISSUE #177: beforeModel only called on nested routes?
Given the following config:

```
this.router.map(function(match) {
  match('/').to('home');
  match('/users').to('users');
  match('/users/:id').to('user');
});
```

If I navigate to `/users/123` the `beforeModel` method is only called for `user`.

However if I set it up like this:

```
match('/').to('home', function(match) {
  match('/users').to('users', function(match) {
    match('/:id').to('user');
  });
});
```

`beforeModel` is called for `/`, `/users`, and `/:id`

In [the section on nesting](https://github.com/tildeio/router.js#nesting-without-handlers) it shows an example of both styles and mentions:

> Both work identically, so do whichever you prefer.

But that doesn't seem to be the case unless I'm missing something? (which is very possible :)

Serabe:
Fixes #177

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,8 @@ makes it easier to change patterns higher up. In this
 case, changing `/posts` to `/pages` would be easier
 in the second example than the first.
 
-Both work identically, so do whichever you prefer.
+Both recognize the same sets of URLs but only the nested
+ones invoke the hooks in the ancestor routes too.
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Awesome, thank you!

PR #200, opened 2016-11-04
Add tests demonstrating that #60 is resolved

pixelhandler:
- Thanks to @jmeas, #164


rwjblue:
Thanks @pixelhandler!

PR #201, opened 2016-11-05
Improve errors

### LINKED ISSUE #174: TransitionAborted should be a subclass of Error, not a POJO
https://github.com/tildeio/router.js/blob/master/lib/router/transition.js#L314-L317

Ember throws `TransitionAborted` objects as errors, but their API is inconsistent with `Error` (e. g. no `stack` property and `.toString()` is ridiculously non-informative).

To resolve this, `TransitionAborted` should be derived from the native `Error` object or one of its [subclasses](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Error_types).

sandstrom:
@rwjblue how does this look?

- Inherit from `Error`
- Cleanup tests to use a shared helper
- Test to ensure that TransitionAbortedError has the correct name
- Improve stacktrace capturing

fixes #174

### DIFF
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -1,5 +1,6 @@
 import { Promise } from 'rsvp';
 import { trigger, slice, log, promiseLabel } from './utils';
+import TransitionAbortedError from './transition-aborted-error';
 
 /**
   A Transition is a thennable (a promise-like object) that represents
@@ -325,16 +326,11 @@ Transition.prototype.send = Transition.prototype.trigger;
 /**
   @private
 
-  Logs and returns a TransitionAborted error.
+  Logs and returns an instance of TransitionAbortedError.
  */
 function logAbort(transition) {
   log(transition.router, transition.sequence, "detected abort.");
-  return new TransitionAborted();
+  return new TransitionAbortedError();
 }
 
-function TransitionAborted(message) {
-  this.message = (message || "TransitionAborted");
-  this.name = "TransitionAborted";
-}
-
-export { Transition, logAbort, TransitionAborted };
+export { Transition, logAbort, TransitionAbortedError as TransitionAborted };
--- a/lib/router/unrecognized-url-error.js
+++ b/lib/router/unrecognized-url-error.js
@@ -1,13 +1,25 @@
 import { oCreate } from './utils';
 
-/**
-  Promise reject reasons passed to promise rejection
-  handlers for failed transitions.
- */
 function UnrecognizedURLError(message) {
-  this.message = (message || "UnrecognizedURLError");
-  this.name = "UnrecognizedURLError";
-  Error.call(this);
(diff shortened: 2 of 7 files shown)


lib/router/unrecognized-url-error.js
```diff
@@ -5,9 +5,25 @@ import { oCreate } from './utils';
   handlers for failed transitions.
  */
```
sandstrom: This description doesn't make any sense to me, should this be replaced with something else?

test/tests/transition-aborted-error_test.js
```diff
@@ -0,0 +1,20 @@
...
+  var error;
+
+  try {
+    throw new TransitionAbortedError('Message');
+  } catch(e) {
+    error = e;
+  }
+
+  // it would be more correct with TransitionAbortedError, but other libraries may rely on this name
+  assert.equal(error.name, 'TransitionAborted', "TransitionAbortedError has the name 'TransitionAborted'");
```
sandstrom: I guess we cannot change the name, since other may rely on this, right?

https://github.com/travis-ci/travis-web/blob/ca5ef0044f3ee4e51c7cf4dd46b1e44e1cc980a8/app/services/raven.js#L7

lolmaus:
Relevant discussion here: #174, has an open question.

test/tests/transition-aborted-error_test.js
nathanhammond: Confirm, not valid to change name.

lib/router/transition.js
```diff
@@ -325,16 +326,11 @@ Transition.prototype.send = Transition.prototype.trigger;
...
+  return new TransitionAbortedError();
 }
 
-function TransitionAborted(message) {
-  this.message = (message || "TransitionAborted");
-  this.name = "TransitionAborted";
-}
-
-export { Transition, logAbort, TransitionAborted };
+export { Transition, logAbort };
```
nathanhammond: We should probably export `TransitionAbortedError` at `TransitionAborted` for back-compat reasons.

rwjblue: Agreed

lib/router/unrecognized-url-error.js
rwjblue: Hehe, I agree it seems pretty whacky.  Mind either removing it completely or adding something that describes reality?

sandstrom:
@rwjblue @nathanhammond I've made the changes. Thanks for taking a look!

test/tests/transition-aborted-error_test.js
rwjblue: FWIW, Ember itself expects `.name` to be `TransitionAborted`: https://github.com/emberjs/ember.js/blob/aeeddf7c00b42a340494465252481b5414d87009/packages/ember-runtime/lib/ext/rsvp.js#L40-L42

PR #202, opened 2016-11-07
Revert "Improve errors"

rwjblue:
Reverts tildeio/router.js#201


rwjblue:
See #203 for details on why we are reverting for now.

@sandstrom - Would you mind resubmitting and addressing that build issue?

sandstrom:
@rwjblue sure, fixed in https://github.com/tildeio/router.js/pull/204

PR #205, opened 2017-01-14
Broken test for queryParam only transitions not using replaceURL

alanwguo:
This is a broken test that demonstrates the issue I am having referenced [here](https://github.com/tildeio/router.js/commit/372364364f61c9a772cf2253093e553248273b97#commitcomment-20461173).

The test does a replaceWith where only the queryParams are changing. This should result in a replaceURL but instead results in a updateURL. 

What seems to happen is that fireQueryParamsDidChange occurs twice. Once in `queryParamsTransition` seen [here](https://github.com/tildeio/router.js/blob/ca5cb2a668431c9fef843235138d4ec7c60dd90d/lib/router/router.js#L122) and once in `getTransitionByIntent ` seen [here](https://github.com/tildeio/router.js/blob/ca5cb2a668431c9fef843235138d4ec7c60dd90d/lib/router/router.js#L83). This causes two refreshes during the replaceWith transition. The second refresh transition aborts the first refresh transition which sets `Transition.isCausedByAbortingTransition` to true.

The router will always use updateURL when `isCausedByAbortingTransition` is true as seen [here](https://github.com/tildeio/router.js/blob/ca5cb2a668431c9fef843235138d4ec7c60dd90d/lib/router/router.js#L645-L654).

### DIFF
--- a/test/tests/router_test.js
+++ b/test/tests/router_test.js
@@ -3985,6 +3985,64 @@ test("Calling transitionTo after initial transition outside validation hook shou
   assert.equal(barModelCount, 1, 'Bar model should be called once');
 });
 
+test("Calling replaceWith after initial transition to change queryParams should use replaceURL", function(assert) {
+  assert.expect(2);
+  map(assert, function(match) {
+    match("/foo").to('foo');
+  });
+
+  var fooHandler = {
+    queryParams: {
+      fooParam: {
+        refreshModel: true,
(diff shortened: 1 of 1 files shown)


rmxdave:
Any update on this? Our team is running into this as well.

wagenet:
Closing due to staleness.

PR #206, opened 2017-02-18
Deps

rmehlinger:
Add explicit dependencies and peer dependencies for rsvp and route-recognizer, per https://codingwithspike.wordpress.com/2016/01/21/dealing-with-the-deprecation-of-peerdependencies-in-npm-3/

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,9 @@
     "grunt-contrib-qunit": "^1.2.0",
     "grunt-s3": "~0.2.0-alpha.2",
     "load-grunt-config": "~0.5.0",
-    "load-grunt-tasks": "~0.2.0"
+    "load-grunt-tasks": "~0.2.0",
+    "route-recognizer": "^0.3.1",
+    "rsvp": "^3.3.3"
   },
   "scripts": {
     "test": "grunt test",
@@ -45,5 +47,9 @@
     "router",
     "route-recognizer",
     "rsvp"
-  ]
+  ],
+  "peerDependencies": {
+    "route-recognizer": "^0.3.0",
+    "rsvp": "^3.0.0"
+  }
 }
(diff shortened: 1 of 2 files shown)


rwjblue:
Not sure what is going on here, but there are quite a few too many commits here.

package.json
```diff
@@ -45,5 +45,13 @@
     "router",
     "route-recognizer",
     "rsvp"
-  ]
+  ],
+  "dependencies": {
```
rwjblue: Should be `devDependencies`

rmehlinger: These are required to run `router.js` from a frontend web application, so it seems like they actually are dependencies and not dev dependencies?

rmehlinger:
Now one commit, as intended.

package.json
rwjblue: The point of `peerDependencies` is that we expect the "host" (the thing that depends upon `route-recognizer`) to provide these dependencies so that it can choose its own version.

We specify `devDependencies` so that our own test suite functions (since in that context it is the "host").

rmehlinger: The issue is, if one wants to use routerjs outside of Ember (which I do, for virtually all of my frontend dev projects), one is unlikely to have route-recognizer or rsvp already installed. It seems then that these are logically dependencies, not peerDependencies. By specifying that these are both dependencies and peerDependencies, we essentially wind up installing rsvp and route-recognizer if they are not already present.

Alternatively, I can drop the dependencies and just add a line to the README about how to install via NPM.

package.json
```diff
@@ -32,6 +32,8 @@
     "grunt-s3": "~0.2.0-alpha.2",
     "load-grunt-config": "~0.5.0",
     "load-grunt-tasks": "~0.2.0"
```
rwjblue: missing a trailing comma here (which makes the JSON invalid)

rwjblue:
Thank you!

PR #210, opened 2017-04-17
Fix query params refresh transitions if they are using replaceState

alexspeller:
So I fixed this in #198. But it was broken again by #197 and the tests didn't
catch that. So I've improved the tests and fixed it again.

@rwjblue I think you were looking at this last time

Ember issue: emberjs/ember.js#11843
Ember failing test PR: emberjs/ember.js#15148


rwjblue:
Thanks @alexspeller!

PR #211, opened 2017-04-25
If a transition initiated with `replaceWith`, it's retry is also a replace.

cibernox:
Prior to this change retrying the transition would convert the `replaceWith` in
a regular `transitionTo`.

Example code (assuming `this` is an Ember route):
```js
let transition = this.replaceWith('new.route');
transition.abort();
session.set('attemptedTransition', transition);
session.authenticate('authenticator:devise').authenticate(username, password); // Successful authentication will call `session.get('attemptedTransition').retry()`
```


rwjblue:
Thanks @cibernox!

PR #213, opened 2017-05-18
Ensure initial transition can be aborted then retried.

rwjblue:
Attempting to recreate the scenario identified in https://github.com/emberjs/ember.js/issues/15190.

### DIFF
--- a/lib/router/transition.js
+++ b/lib/router/transition.js
@@ -235,7 +235,18 @@ Transition.prototype = {
     // TODO: add tests for merged state retry()s
     this.abort();
     var newTransition = this.router.transitionByIntent(this.intent, false);
-    newTransition.method(this.urlMethod);
+
+    // inheriting a `null` urlMethod is not valid
+    // the urlMethod is only set to `null` when
+    // the transition is initiated *after* the url
+    // has been updated (i.e. `router.handleURL`)
+    //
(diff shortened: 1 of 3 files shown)


rwjblue:
@alexspeller @cibernox - This fixes https://github.com/emberjs/ember.js/issues/15190 (if y'all are curious to the root cause).

lib/router/transition.js
```diff
@@ -235,7 +235,18 @@ Transition.prototype = {
...
+    // inheriting a `null` urlMethod is not valid
+    // the urlMethod is only set to `null` when
+    // the transition is initiated *after* the url
+    // has been updated (i.e. `router.handleURL`)
+    //
+    // in that scenario, the url method cannot be
+    // inherited for a new transition because then
+    // the url would not update even though it should
+    if (this.urlMethod !== null) {
+      newTransition.method(this.urlMethod);
```
cibernox: I'm a bit confused actually. I thought that the `urlMethod` could be either `update` or `replace`. Under what circumstances is it `null`?

rwjblue: FWIW, I think it may be better to just add another flag for this to the Transition object.  That way the `urlMethod` doesn't need to play double duty, and its much more obvious what is going on...

PR #214, opened 2017-05-22
Bind the serialize handler to the correct "this"

boris-petrov:
I hit this issue in an Ember application where `this` in the `serialize` hook inside a router was undefined. I guess this is a bug.

Not sure whether the same should be done for `this.serializer` on the line that I've changed. I'm also not sure that I've handled the `async` case in the tests correctly - I just ignore it.

I'm willing to fix anything that you don't like in this PR. :)

### DIFF
--- a/lib/router/handler-info/unresolved-handler-info-by-object.js
+++ b/lib/router/handler-info/unresolved-handler-info-by-object.js
@@ -24,8 +24,7 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
   */
   serialize: function(_model) {
     var model = _model || this.context,
-      names = this.names,
-      serializer = this.serializer || (this.handler && this.handler.serialize);
+      names = this.names;
 
     var object = {};
     if (isParam(model)) {
@@ -34,8 +33,11 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
     }
 
     // Use custom serialize if it exists.
-    if (serializer) {
-      return serializer(model, names);
+    if (this.serializer) {
+      // invoke this.serializer unbound (getSerializer returns a stateless function)
+      return this.serializer.call(null, model, names);
+    } else if (this.handler && this.handler.serialize) {
+      return this.handler.serialize(model, names);
     }
 
     if (names.length !== 1) {
--- a/tests/router_test.js
+++ b/tests/router_test.js
@@ -1491,6 +1491,22 @@ scenarios.forEach(function(scenario) {
     );
   });
 
+  test('the serializer method is unbound', function(assert) {
+    assert.expect(1);
+
+    router.getSerializer = function() {
+      return function(date) {
+        assert.equal(this, undefined);
+        return {
+          date:
+            date.getFullYear() + '-' + date.getMonth() + '-' + date.getDate(),
+        };
+      };
+    };
+
+    router.generate('showPostsForDate', new Date(1815, 5, 18));
+  });
+
   test('params are known by a transition up front', function(assert) {
     assert.expect(2);
 
--- a/tests/router_test.js
+++ b/tests/router_test.js
@@ -2925,6 +2925,26 @@ scenarios.forEach(function(scenario) {
       });
   });
 
+  test('the serialize function is bound to the correct object when called', function(
+    assert
+  ) {
+    assert.expect(scenario.async ? 0 : 1);
+
+    handlers = {
+      showPostsForDate: {
+        serialize: function(date) {
+          assert.equal(this, handlers.showPostsForDate);
+          return {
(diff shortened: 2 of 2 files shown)


lib/router/handler-info/unresolved-handler-info-by-object.js
```diff
@@ -25,7 +25,7 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
   serialize: function(_model) {
     var model = _model || this.context,
         names = this.names,
-        serializer = this.serializer || (this.handler && this.handler.serialize);
+        serializer = this.serializer || (this.handler && this.handler.serialize && this.handler.serialize.bind(this.handler));
```
stefanpenner: how does this get unbound?

boris-petrov: That's the way JavaScript works. When you get a reference to a method, `this` in it is not bound. The same applies for `this.serializer`, but I couldn't figure out where you use/set that and that's why I left it unmodified.

stefanpenner: Oh sorry, i totally didn't read the code correctly. You are correct!

stefanpenner: I should not be allowed to comment pre-coffee...

stefanpenner: @trentmwillis it looks like you added this, was this intentional to decoupling serialize from the `this` re: lazy engines? Or an oversight?

stefanpenner: We should add a test regardless, ensuring behavior remains as we want.

stefanpenner:
> this.serializer is intentionally unbound though.

We should then likely change this PR to test `this.serialize` is unbound.

boris-petrov:
I pushed another commit with another test to check that `this.serializer` is unbound. Please let me know if you need any more changes.

spectras:
I just stumbled upon this. For what it's worth, invoking route's `serialize` method with no `this` breaks the use of mixins in Ember routes, as `superWrapper` looks for `this._super`.

It also prevents having a base route object with some generic methods that you can extend and just set a few attributes.

Minimalist example hitting both:
```js
// my-mixin.js
export default Ember.Mixin.create({
    serialize: function (model, params) {
        const key = get(this, 'modelName') + '_id';
        if (get(model, 'isNew')) {
            let result = {};
            result[key] = 'new';
            return result;
        }
        return this._super(model, params);
    }
});
```
```js
// my-route.js
export default Ember.Route.extend(MyMixin, {
    modelName: 'myModel'
});
```

lib/router/handler-info/unresolved-handler-info-by-object.js
```diff
@@ -25,7 +25,8 @@ var UnresolvedHandlerInfoByObject = subclass(HandlerInfo, {
   serialize: function(_model) {
     var model = _model || this.context,
         names = this.names,
-        serializer = this.serializer || (this.handler && this.handler.serialize);
+        // this.serializer is intentionally unbound since getSerializer is expected to return a stateless function
+        serializer = this.serializer || (this.handler && this.handler.serialize && this.handler.serialize.bind(this.handler));
```
rwjblue: I would prefer to refactor this a bit to avoid needing to `.bind` again for every call to the same handler's `serialize`.  I think the simplest way woul dbe to remove this `serializer` local variable, and update the `if` conditional below to:

```js
if (this.serializer) {
  // invoke this.serializer unbound (getSerializer returns a stateless function)
  return (0, this.serializer)(model, names);
} else if (this.handler && this.handler.serialize) {
  return this.handler.serialize(model, names);
}
```

boris-petrov:
@rwjblue - I did the changes you requested, you're right that it is much better that way. However there's a JSHint error which I'm not sure how to resolve:

`router/handler-info/unresolved-handler-info-by-object.js: line 38, col 34, Bad invocation.`

This is on the line:

```javascript
return (0, this.serializer)(model, names);
```

boris-petrov:
@trentmwillis - thanks for the tip, I used the `.call` version - I prefer it.

@rwjblue - I did the changes as requested, rebased on master and fixed the tests. Please merge this soon as I don't want to rebase again after a million changes. :D

Turbo87:
FWIW the following workaround in the relevant routes seems to work fine if you're stuck on older Ember versions:

```js
  init() {
    this._super(...arguments);

    // see https://github.com/tildeio/router.js/pull/214
    this.serialize = this.serialize.bind(this);
  },
```

rwjblue:
Thank you so much @boris-petrov!

PR #215, opened 2017-05-29
using `Object.create(null)` for resolvedModels

bekzod:
will make corresponding changes to ember.js once merged like removing `transition.resolvedModels.hasOwnProperty`


rwjblue:
What is the purpose of the change? Does it give us anything specific?

bekzod:
closing this old PR don't remember why I did this

PR #224, opened 2017-09-08
Update to more modern syntax.

rwjblue:
* Use re-exports for entry point
* Convert `TransitionIntent` and `HandlerInfo`'s to ES classes
* Modify `lib/router/utils.js`
  * Use const/let throughout
  * Remove `Object.create` polyfill
  * Remove `Array.isArray` polyfill
  * Remove `bind` helper in favor of using `Function.prototype.bind`
  * Remove `subclass` utility method


lib/router/handler-info.js
```diff
@@ -238,40 +227,54 @@ HandlerInfo.prototype = {
...
+// that the params _must_ match also in that case.
+//
+// The only reason `oldHandlerInfo.context` and `newHandlerInfo.context` did not
+// match in prior versions is because if the context isn't set yet (on newHandlerInfo)
+// is because it inherits the `null` from the prototype vs `undefined` (on
+// the oldHandlerInfo).
+//
+// A future refactoring should remove that conditional, and fix the hand full of
+// failing tests.
+HandlerInfo.prototype.context = null;
```
rwjblue: Made an issue to track this future cleanup/fixing here: https://github.com/tildeio/router.js/issues/225.

PR #230, opened 2017-10-04
convert Router to es6 class

bekzod:


rwjblue:
@bekzod - Just noticed a few more `var`'s inside `getTransitionByIntent`(moved to the bottom of the diff), might be fodder for another PR...

PR #231, opened 2017-10-04
update `rsvp` and `route-recognizer`

bekzod:


bekzod:
`yarn.lock` updated

rwjblue:
Thank you!

PR #232, opened 2017-10-04
Es6 cleanup

bekzod:


lib/router/handler-info.js
```diff
@@ -72,46 +69,36 @@ export default class HandlerInfo {
   }
 
   resolve(shouldContinue, payload) {
-    var checkForAbort = this.checkForAbort.bind(this, shouldContinue),
-      beforeModel = this.runBeforeModelHook.bind(this, payload),
-      model = this.getModel.bind(this, payload),
-      afterModel = this.runAfterModelHook.bind(this, payload),
-      becomeResolved = this.becomeResolved.bind(this, payload),
-      self = this;
+    let checkForAbort = this.checkForAbort.bind(this, shouldContinue);
```
locks: 😍

lib/router/handler-info.js
```diff
@@ -72,46 +69,36 @@ export default class HandlerInfo {
...
+    let beforeModel = this.runBeforeModelHook.bind(this, payload);
+    let model = this.getModel.bind(this, payload);
+    let afterModel = this.runAfterModelHook.bind(this, payload);
+    let becomeResolved = this.becomeResolved.bind(this, payload);
 
     return Promise.resolve(
       this.handlerPromise,
       this.promiseLabel('Start handler')
-    ).then(
-      function(handler) {
```
rwjblue: This is the only part that concerns me (generally speaking).  The changes here seem correct, but I need to dig into what this comment was trying to say before we merge.

bekzod: looks like weird promise chain optimization, for me it should work same as before :P

PR #237, opened 2017-11-01
cleanup `extractQueryParams`

bekzod:


lib/router/utils.js
```diff
@@ -25,19 +25,9 @@ export function merge(hash, other) {
...
-  if (
-    len &&
-    len > 0 &&
-    array[len - 1] &&
-    array[len - 1].hasOwnProperty('queryParams')
-  ) {
-    queryParams = array[len - 1].queryParams;
-    head = slice.call(array, 0, len - 1);
-    return [head, queryParams];
+  let possibleQuery = array[array.length - 1];
```
rwjblue: The prior version supported `array` being `undefined`, and this one doesn't.  Are we 100% certain that all callsites pass a valid array?

lib/router/utils.js
```diff
@@ -25,19 +25,9 @@ export function merge(hash, other) {
...
+  if (possibleQuery && possibleQuery.hasOwnProperty('queryParams')) {
```
rwjblue: I'd prefer to change this to `'queryParams' in possibleQuery` (or use `Object.prototype.hasOwnProperty.call(possibleQuery, 'queryParams')`).

lib/router/utils.js
rwjblue: Also, isn't this actually `possibleOptions` (since the "query params" object is nested inside of it)?

wagenet:
Closing due to staleness.

PR #240, opened 2018-02-13
When calling replace on a replace, it should replace not update

wagenet:


lib/router/router.js
```diff
@@ -769,7 +769,21 @@ function updateURL(transition, state /*, inputUrl*/) {
     var isQueryParamsRefreshTransition =
       transition.queryParamsOnly && urlMethod === 'replace';
 
-    if (initial || replaceAndNotAborting || isQueryParamsRefreshTransition) {
+    // say you are at / and you a `replaceWith(/foo)` is called. Then, that
+    // transition is aborted with `replaceWith(/bar)`. At the end, we should
+    // end up with /bar replacing /. We are replacing the replace. We only
+    // will replace the initial route if all subsequent aborts are also
+    // replaces. However, there is some ambiguity around the correct behavior
```
rwjblue: Can you describe the ambiguity you are referring to here? (I'd like to make sure we are all on the same page RE: the factors and assumptions)

tests/router_test.js
```diff
@@ -4595,6 +4596,84 @@ test("A failed handler's setup shouldn't prevent future transitions", function(a
     assert.equal(barModelCount, 1, 'Bar model should be called once');
   });
 
+  test('Calling replaceWith after initial replace in validation hook with multiple redirects should use replaceUrl', function(assert) {
+    // assert.expect(8);
```
rwjblue: Why is this commented?

tests/router_test.js
```diff
@@ -4595,6 +4596,84 @@ test("A failed handler's setup shouldn't prevent future transitions", function(a
...
+      url = updateUrl;
+      history.push(url);
+    };
+
+    router.replaceURL = function(replaceURL) {
+      url = replaceURL;
+      if (history.length > 0) {
+        history[history.length - 1] = url;
+      } else {
+        history.push(url);
```
rwjblue: Is this actually correct? My understanding of the current solution to "redirects during initial transition" is that we _will not_ call replaceURL prior to updateURL. If that is still correct then this conditional should not be needed...

lib/router/router.js
wagenet: The ambiguity I see is a replace followed by a series that includes an update. It's not clear to me if that should add a history entry or not.

tests/router_test.js
wagenet: I commented it out since I kept changing the number of asserts. Should probably just remove it since I'm not doing any asserts in callbacks.

tests/router_test.js
wagenet: TBH, I wasn't thinking that carefully about that bit. I can remove it and see how it goes.

tests/router_test.js
rwjblue: Ya, makes sense

wagenet:
@rwjblue wouldn't that other test be: https://github.com/wagenet/router.js/blob/55a4d59c237e330f01afec57872f510e4be5851e/tests/router_test.js#L4677 ?

tests/router_test.js
rwjblue: Yep, sounds good to me.

wycats:
@rwjblue FYI we hit this bug in our app, resulting in the "back button of doom" bug; I wonder whether it happens a lot and people just don't think to report it?

rwjblue:
Ya, I haven’t seen it reported in various repos I’ve been following. Seems like a good fix to me. I’m glad @wagenet tracked it down...

wagenet:
I think we were more susceptible because we were doing some other weird things (as a result of code written 4 years ago).

@rwjblue anything else you want here?

Issue #236, opened 2017-10-30
param-only contexts are improperly considered equal when params differ


bgentry:
There's a long-running Ember issue (https://github.com/emberjs/ember.js/issues/12419) that discusses a problem where, when using `link-to` with a model's ID instead of the model itself, all links to the same route get the `.ember-transitioning-in` class applied to them, even though their params are different.

I just dug very deep into what was causing this when I encountered it on my app. I wrote up a detailed summary of my findings in this comment, so rather than re-post it here I'll just link to it: https://github.com/emberjs/ember.js/issues/12419#issuecomment-340341755

In short, [`isActiveIntent`](https://github.com/tildeio/router.js/blob/0235c1de0aeb2c6c44215f7ab3ce56e8f4651227/lib/router/router.js#L301) incorrectly returns true in this situation because it is not correctly using the `contexts` that are provided to it when checking route equality.

Though I was able to dig this far, I really have no idea what the appropriate fix is. Hopefully with this information it will be straightforward for somebody who understands router.js.

Cheers :v:

rwjblue:
should have been addressed by #242

PR #242, opened 2018-02-24
Prevent context matching if the oldContext is null

nathanhammond:
Fixes:
- https://github.com/emberjs/ember.js/issues/12419
- https://github.com/emberjs/ember.js/issues/10989
- https://github.com/emberjs/ember.js/issues/10195
- https://github.com/tildeio/router.js/issues/236
- https://github.com/emberjs/ember.js/issues/16275

### DIFF
--- a/lib/router/transition-intent/named-transition-intent.js
+++ b/lib/router/transition-intent/named-transition-intent.js
@@ -114,7 +114,11 @@ export default class NamedTransitionIntent extends TransitionIntent {
           newHandlerInfo.context
         );
         var oldContext = oldHandlerInfo && oldHandlerInfo.context;
-        if (result.names.length > 0 && newHandlerInfo.context === oldContext) {
+        if (
+          result.names.length > 0 &&
+          oldHandlerInfo.hasOwnProperty('context') &&
+          newHandlerInfo.context === oldContext
+        ) {
           // If contexts match in isActive test, assume params also match.
           // This allows for flexibility in not requiring that every last
           // handler provide a `serialize` method
--- a/lib/router/transition-intent/named-transition-intent.js
+++ b/lib/router/transition-intent/named-transition-intent.js
@@ -116,7 +116,7 @@ export default class NamedTransitionIntent extends TransitionIntent {
         var oldContext = oldHandlerInfo && oldHandlerInfo.context;
         if (
           result.names.length > 0 &&
-          oldHandlerInfo.hasOwnProperty('context') &&
+          'context' in oldHandlerInfo &&
           newHandlerInfo.context === oldContext
         ) {
           // If contexts match in isActive test, assume params also match.
--- a/tests/router_test.js
+++ b/tests/router_test.js
@@ -1549,6 +1549,48 @@ scenarios.forEach(function(scenario) {
     transitionTo(router, 'adminPost', { id: 75 });
   });
 
+  test('check for mid-transition correctness', function(assert) {
+    var posts = {
+      1: { id: 1 },
+      2: { id: 2 },
+      3: { id: 3 },
+    };
+
+    var showPostHandler = {
+      serialize: function(object) {
+        return (object && { id: object.id }) || null;
+      },
+
+      model: function(params) {
+        return posts[params.id];
+      },
+    };
+
+    handlers = {
+      showPost: showPostHandler,
+    };
+
+    // Get a reference to the transition, mid-transition.
+    router.willTransition = function() {
+      var midTransitionState = router.activeTransition.state;
+
+      // Make sure that the activeIntent doesn't match post 300.
+      var isPost300Targeted = router.isActiveIntent(
+        'showPost',
+        [300],
+        null,
+        midTransitionState
+      );
+      assert.notOk(isPost300Targeted, 'Post 300 should not match post 3.');
+    };
+
+    // Go to post 3. This triggers our test.
+    transitionTo(router, '/posts/3');
+
+    // Clean up.
+    delete router.willTransition;
+  });
+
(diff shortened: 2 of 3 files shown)


rwjblue:
Why is it not possible to test (can you explain)?

nathanhammond:
To trigger the issue the route actually needs to be processed asynchronously. The faked-in asynchronous behavior isn't enough to trigger it.

So, I would need to [modify this line](https://github.com/tildeio/router.js/blob/5b08deb39d1d22f4eeb0a943fb2d7b3f6518863b/tests/router_test.js#L31) like so:

```diff
diff --git a/tests/router_test.js b/tests/router_test.js
index 07b2d69..cb3bafd 100644
--- a/tests/router_test.js
+++ b/tests/router_test.js
@@ -28,7 +28,15 @@ var scenarios = [
     getHandler: function(name) {
       // Treat 'loading' route transitions are synchronous
       var handler = handlers[name] || (handlers[name] = {});
-      return name === 'loading' ? handler : Promise.resolve(handler);
+      if (name === 'loading') {
+        return handler;
+      } else {
+        return new Promise(function(resolve) {
+          setTimeout(function() {
+            resolve(handler);
+          }, 1);
+        });
+      }
     },
     getSerializer: function(name) {
       return serializers && serializers[name];
```

And then approximately every async test fails since it is now running out of turn compared to the test suite.

nathanhammond:
(I mean, I can create a new test harness just for this, I suppose.)

nathanhammond:
Also, I just found this gem:
https://github.com/tildeio/router.js/blame/c94df4be011665fcf3a4ad464413e927a9c648d3/lib/router/handler-info.js#L240-L254

nathanhammond:
The problem with the old code is that it was simply comparing equality. We really need it to check to see if the contexts were matching and explicitly set. If so, it's a valid assumption to copy over params.

lib/router/transition-intent/named-transition-intent.js
```diff
@@ -114,7 +114,11 @@ export default class NamedTransitionIntent extends TransitionIntent {
           newHandlerInfo.context
         );
         var oldContext = oldHandlerInfo && oldHandlerInfo.context;
-        if (result.names.length > 0 && newHandlerInfo.context === oldContext) {
+        if (
+          result.names.length > 0 &&
+          !!oldContext &&
```
rwjblue: We should be specific here I think. For example, a `0` or `’’` (empty string) are probably fine...

nathanhammond:
The bug is triggered by this process:

1. Generate a new `TransitionState`.
2. Start that `TransitionState` becoming resolved.
3. Before that transition has resolved (thus the async requirement for failure)...
4. ... attempt to check if the destination route is active using `isActiveIntent`.
5. Feed in the partially-completed transition to the "private" `_state` argument so that we compare against the ongoing transition.
6. This will trigger an attempt to read the `context` on `oldHandlerInfo` before it is assigned.
7. Then, because `oldHandlerInfo` (which isn't even finished being processed yet) doesn't have a `context`, it assigns `oldContext` to `null`.
8. This matches the prototype value from `newHandlerInfo` (obviously, sharing a prototype).
9. Then we clobber `newHandlerInfo`'s `params` because we didn't count on this race condition.

nathanhammond:
(Travis lost its mind. Restart the failing test one to see it actually fail.)

rwjblue:
Restarted a few times, seems to not fail?

lib/router/transition-intent/named-transition-intent.js
```diff
@@ -114,7 +114,11 @@ export default class NamedTransitionIntent extends TransitionIntent {
           newHandlerInfo.context
         );
         var oldContext = oldHandlerInfo && oldHandlerInfo.context;
-        if (result.names.length > 0 && newHandlerInfo.context === oldContext) {
+        if (
+          result.names.length > 0 &&
+          oldHandlerInfo.hasOwnProperty('context') &&
```
rwjblue: Does adding this guard allow us to delete this https://github.com/tildeio/router.js/blob/2bfd8e5484f6925d0010777a000782cd84d701ca/lib/router/handler-info.js#L240-L254

If we can do that, we don't need the `hasOwnProperty` check here...

tests/router_test.js
```diff
@@ -1594,6 +1594,27 @@ scenarios.forEach(function(scenario) {
       showPost: showPostHandler,
     };
 
+    // Check for mid-transition correctness.
```
rwjblue: Can you move this into a test of its own? Seems odd to just be an extra assertion in a somewhat unrelated test...

lib/router/transition-intent/named-transition-intent.js
nathanhammond: We're still going to need `hasOwnProperty` for the scenario where: `oldHandlerInfo.context === undefined && newHandlerInfo.context === undefined` because neither explicitly set it.

It's also theoretically possible that use as a microlib would set context to a "meaningful" `undefined` value on the instance.

Regardless, I believe that `hasOwnProperty` is the correct check here.

nathanhammond: (See step 7 in https://github.com/tildeio/router.js/pull/242#issuecomment-368662393)

nathanhammond:
- Test extracted.
- Removal of the "default" context from the prototype is added as an optional separate commit.

I opted to never explicitly set the `context` on a handler unless we actually received a `context`. Since we always compare identity but could have a primitive show up (`undefined`) we need to use `hasOwnProperty` to distinguish between an `undefined` `context` and an `unset` `context`.

PR #243, opened 2018-02-27
Failing test for mid-transition active checks.

nathanhammond:
Demonstrate test failure because Travis has lost its mind.


nathanhammond:
(Use the original PR, which now shares the same cherry-picked hashes from here.)

rwjblue:
lol, wtf

PR #245, opened 2018-04-28
Fix a crash when (hash) helper is used to create queryParams object

tchak:
fix `{{url-for 'posts' (hash queryParams=paramsHash)}}` use case


tchak:
Might be fixed by #237 in which case feel free to close

rwjblue:
Thanks @tchak!

PR #252, opened 2018-09-18
[Breaking] RouteInfo Type And Transition.(from|to)

chadhietala:
This introduces the public `RouteInfo` type and `from` and `to` fields on the `Transition` as described in https://github.com/emberjs/rfcs/pull/95.

This PR also:
- [Breaking] removes all underscore methods from the `Route` type.
- [Breaking] Renames `HandlerInfo` -> `RouteInfo` in many places.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- 6
+  - 6
 
 sudo: false
 dist: trusty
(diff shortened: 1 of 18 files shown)


rwjblue:
How do you plan to mitigate the breakage on the Ember side when we land this there?

chadhietala:
@rwjblue all of this should absorbed by Ember. With the private infos the only thing that might likely troll users is that I renamed the `handle` property to `route`, but other than that the structure should be the same.

ef4:
This will likely break liquid-outlet. But that’s fine, the whole point was creating public API to replace the private one it’s using

PR #257, opened 2018-09-27
Add routeWillChange, routeDidChange, transitionDidError

chadhietala:
This introduces two new event types `routeWillChange` and `routeDidChange`. These events are specified by the [Router Service RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#new-events-routewillchange--routedidchange). Unlike the `willTransition` and `didTransition` hooks the new events see all transition events including aborts and errors.

In the case of the error we have always aborted the transition. However, the mechanics here are somewhat problematic since calling `abort` will result in a `routeWillChange` event. To deal with this reality, we have decoupled `abort` from the functionality that rolled back the state of the transition. We have also forced the implementer to handle the transition error. In doing so the host can decide how to handle the error and if it should beacon the events.

In practice, Ember will likely just `rollback` the transition if there is an error substate and `abort` if thee isn't a substate.

### DIFF
--- a/lib/router/route-info.ts
+++ b/lib/router/route-info.ts
@@ -1,5 +1,5 @@
 import { Promise } from 'rsvp';
-import { Dict, Maybe, Option } from './core';
+import { Dict, Option } from './core';
 import Router, { SerializerFunc } from './router';
 import InternalTransition, {
   isTransition,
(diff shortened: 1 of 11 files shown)


chadhietala:
@asakusuma So I added cases for the first 2. The third case is not valid, any redirection that occurs is going to cause the `routeWillChange` to be called. In those cases you will get more `routeWillChange` events than `routeDidChange` events.

asakusuma:
@chadhietala sorry, I meant before `routeDidChange` is called. So basically a test with multiple `transitionTo/replaceWith` calls before the `didTransition` equivalent

lib/router/route-info.ts
```diff
@@ -37,29 +37,40 @@ export type Continuation = () => PromiseLike<boolean> | boolean;
...
   find(
-    predicate: (this: void, routeInfo: RouteInfo, i: number) => boolean,
-    thisArg: any
+    predicate: (this: any, routeInfo: RouteInfo, i?: number) => boolean,
+    thisArg?: any
   ): RouteInfo | undefined;
 }
 
-let ROUTE_INFO_LINKS = new WeakMap<InternalRouteInfo<Route>, RouteInfo>();
+let INTERNAL_TO_PUBLIC = new WeakMap<InternalRouteInfo<Route>, RouteInfo>();
```
rwjblue: This variable name confuses me...

lib/router/route-info.ts
```diff
@@ -37,29 +37,40 @@ export type Continuation = () => PromiseLike<boolean> | boolean;
...
 
-export function toReadOnlyRouteInfo(routeInfos: InternalRouteInfo<Route>[]) {
+export function toReadOnlyRouteInfo(
+  routeInfos: InternalRouteInfo<Route>[],
+  queryParams: Dict<unknown> = {}
+) {
   return routeInfos.map((info, i) => {
-    let { name, params, queryParams, paramNames } = info;
+    let { name, params, paramNames } = info;
     let publicRouteInfo = new class implements RouteInfo {
```
rwjblue: Is there a particular reason this class has to be generated each time `toReadOnlyRouteInfo` is called?

lib/router/transition.ts
```diff
@@ -239,19 +233,30 @@ export default class Transition<T extends Route> implements Partial<Promise<T>>
...
-    }
-    log(this.router, this.sequence, this.targetName + ': transition was aborted');
-    if (this.intent !== undefined && this.intent !== null) {
-      this.intent.preTransitionState = this.router.state;
-    }
-    this.isAborted = true;
-    this.isActive = false;
-    this.router.activeTransition = undefined;
+    this.rollback();
+    this.to = this.from;
```
rwjblue: This seems slightly odd to me. Don't we want to essentially model `Transition`'s as immutable (at least their from/to info) and create _new_ transition instances when we want to mutate?

lib/router/route-info.ts
chadhietala: One would think you could use the `WeakMap` as a cache, however there seems to be some mutation going on related to QPs which results in a corrupted cache.

chadhietala: After further investigation, this does not work without significant work around mutability of ResolvedRouteInfos.

PR #261, opened 2018-10-02
Add recognize and recognizeAndLoad

chadhietala:
This adds 2 new APIs that were defined in https://github.com/emberjs/rfcs/pull/95.

- `recognize`: Takes a string URL and returns a `RouteInfo` for the leafmost route represented by the URL. Returns null if the URL is not recognized. This method expects to receive the actual URL as seen by the browser including the app's `rootURL`.

- `recognizeAndLoad`: Takes a string URL and returns a promise that resolves to a `RouteInfoWithAttributes` for the leafmost route represented by the URL. The promise rejects if the URL is not recognized or an unhandled exception is encountered. This method expects to receive the actual URL as seen by the browser including the app's rootURL.

### DIFF
--- a/lib/router/route-info.ts
+++ b/lib/router/route-info.ts
@@ -49,15 +49,20 @@ export interface RouteInfo {
   ): RouteInfo | undefined;
 }
 
+export interface RouteInfoWithAttributes extends RouteInfo {
+  attributes: Dict<unknown>;
+}
+
 let ROUTE_INFOS = new WeakMap<InternalRouteInfo<Route>, RouteInfo>();
 
 export function toReadOnlyRouteInfo(
   routeInfos: InternalRouteInfo<Route>[],
-  queryParams: Dict<unknown> = {}
+  queryParams: Dict<unknown> = {},
+  includeAttributes = false
 ) {
   return routeInfos.map((info, i) => {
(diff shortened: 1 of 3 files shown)


lib/router/route-info.ts
```diff
@@ -49,15 +49,20 @@ export interface RouteInfo {
...
   routeInfos: InternalRouteInfo<Route>[],
-  queryParams: Dict<unknown> = {}
+  queryParams: Dict<unknown> = {},
+  includeAttributes = false
 ) {
   return routeInfos.map((info, i) => {
-    let { name, params, paramNames } = info;
-    let publicRouteInfo = new class implements RouteInfo {
+    let { name, params, paramNames, context } = info;
+    let RoutInfoImpl = class RoutInfoImpl {
```
rwjblue: Do we actually need a class here? Seems like we could just use a POJO...

chadhietala: If I do `Object.assign` is it going to get re-written to something in Ember?

chadhietala:
I just copy and pasted these definitions from the RFC. `rootURL` is an Ember thing and not a router.js thing. I suspect that we need to triage this in integration on the Ember side.

PR #269, opened 2018-11-06
[BUGFIX] Fix mutable QPs through evented hooks

chadhietala:
After doing some investigation we really need to rationalization QP finalization. The reason why saving off the Route Recognizer QPs is required is because `finalizeQueryParamChange` basically is a [validation step](https://github.com/tildeio/router.js/blob/master/lib/router/router.ts#L728) from the host. In the case of Ember, only QPs that are on the controller are [considered valid](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/routing/lib/system/route.ts#L2418). Unless the "valid" QPs are returned we will effectively return a new empty object [here](https://github.com/tildeio/router.js/blob/master/lib/router/router.ts#L738-L746). This then mutates the [old transition state](https://github.com/tildeio/router.js/blob/master/lib/router/router.ts#L125-L129) in this case clobbers the real QPs with an empty object.

The `RouteInfo` should not be concerned with Ember's validation of QPs.

### DIFF
--- a/lib/router/router.ts
+++ b/lib/router/router.ts
@@ -39,6 +39,7 @@ export interface ParsedHandler {
 }
 
 export default abstract class Router<T extends Route> {
+  private _lastQueryParams = {};
   log?: (message: string) => void;
   state?: TransitionState<T> = undefined;
   oldState: Maybe<TransitionState<T>> = undefined;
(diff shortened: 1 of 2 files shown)


lib/router/router.ts
```diff
@@ -46,6 +46,7 @@ export default abstract class Router<T extends Route> {
   currentRouteInfos?: InternalRouteInfo<T>[] = undefined;
   _changedQueryParams?: Dict<unknown> = undefined;
   currentSequence = 0;
+  lastQueryParams = {};
```
rwjblue: Should probably be marked as private (and underscored)

chadhietala:
Going to add more tests around this stuff.

PR #275, opened 2018-12-05
[FEATURE] buildRouteInfoMetadata

chadhietala:
This is the router.js work required for https://github.com/emberjs/rfcs/pull/398

### DIFF
--- a/lib/router/route-info.ts
+++ b/lib/router/route-info.ts
@@ -33,6 +33,7 @@ export interface Route {
   _internalReset?(wasReset: boolean, transition?: Transition): void;
   contextDidChange?(): void;
   redirect?(context: Dict<unknown>, transition: Transition): void;
+  buildRouteInfoMetadata?(): unknown;
 }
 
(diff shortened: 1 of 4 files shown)


lib/router/route-info.ts
```diff
@@ -63,9 +65,10 @@ export function toReadOnlyRouteInfo(
   includeAttributes = false
 ): RouteInfoWithAttributes[] | RouteInfo[] {
   return routeInfos.map((info, i) => {
-    let { name, params, paramNames, context } = info;
+    let { name, params, paramNames, context, route } = info;
     if (ROUTE_INFOS.has(info) && includeAttributes) {
       let routeInfo = ROUTE_INFOS.get(info)!;
+      routeInfo = attachMetadata(route!, routeInfo);
```
tomdale: Why do we need to assert `route` is non-null here? Is it guaranteed to be populated by this phase of the lifecycle?

chadhietala: Yes, it is in a resolved state. This function is also used for the pre-resolved state.

PR #283, opened 2019-03-22
Make missing parameter error more descriptive

mariokostelac:
While debugging some routing errors on deeply nested routes with many parameters, it became difficult to track what parameters were missing. This collects all missing parameters and adds that to the error message.

### DIFF
--- a/lib/router/transition-intent/named-transition-intent.ts
+++ b/lib/router/transition-intent/named-transition-intent.ts
@@ -206,6 +206,7 @@ export default class NamedTransitionIntent<T extends Route> extends TransitionIn
 
     // Soak up all the provided string/numbers
     let numNames = names.length;
+    let missingParams = [];
     while (numNames--) {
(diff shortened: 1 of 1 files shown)


lib/router/transition-intent/named-transition-intent.ts
```diff
@@ -222,13 +223,16 @@ export default class NamedTransitionIntent<T extends Route> extends TransitionIn
...
-              name
-          );
+          missingParams.push(paramName);
         }
       }
     }
+    if (missingParams.length > 0) {
+      throw new Error(
+        "You didn't provide enough string/numeric parameters to satisfy all of the dynamic segments for route ${name}." +
+          ' Missing params: ${missingParams}'
```
mariokostelac: unfortunately linter thinks it's a good idea to put this line with single quotes even though previous one is double quoted.

PR #288, opened 2019-09-15
Upgrade ember-cli, linting deps, etc

xg-wang:
ci is using node 6 and several packages were already requiring node 6 so I guess this is fine...

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,9 @@
     "broccoli-babel-transpiler": "^6.1.2",
     "broccoli-concat": "^3.7.4",
     "broccoli-funnel": "^2.0.2",
-    "broccoli-merge-trees": "^2.0.0",
+    "broccoli-merge-trees": "^3.0.2",
     "broccoli-typescript-compiler": "^4.0.1",
-    "ember-cli": "^3.12.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -18,33 +21,34 @@
...
+    "prettier": "1.18.2",
+    "qunit": "^2.9.2",
     "route-recognizer": "^0.3.4",
-    "rsvp": "^4.8.1",
-    "tslint": "^5.11.0",
-    "tslint-config-prettier": "^1.14.0",
-    "tslint-plugin-prettier": "^1.3.0",
-    "typescript": "~3.0.1"
+    "rsvp": "^4.8.5",
+    "tslint": "^5.20.0",
```
rwjblue: FWIW, we should migrate to eslint (since tslint is deprecated) but we can do it in a separate PR...

package.json
```diff
@@ -54,4 +58,4 @@
     "node": "^4.5 || 6.* || >= 8.*"
```
rwjblue: We should probably also bump this to:

```suggestion
    "node": "8.* || 10.* || >= 12.*"
```

PR #289, opened 2019-11-04
Upgrade typescript

xg-wang:
Seems there's still some incorrect typing need to be fixed.


xg-wang:
@rwjblue this PR is a bit outdated, since the latest broccoli-typescript-compiler is already v6

Issue #291, opened 2020-01-17
Can't add property attributes, object is not extensible


raido:
I am hitting this error with Ember.js v3.13+:

![Screenshot 2020-01-17 at 13 29 12](https://user-images.githubusercontent.com/659945/72610854-2da21880-3931-11ea-92e9-1f93708f346d.png)

This is happening on older browsers like Android 5 with Chrome v37 etc.

My app also includes Babel polyfill.

--
Here's my `targets.js` setup:

```
'use strict';

const browsers = [
  'last 1 Chrome versions',
  'last 1 Firefox versions',
  'last 1 Safari versions'
];

const isCI = !!process.env.CI;
const isProduction = process.env.EMBER_ENV === 'production';

if (isCI || isProduction) {
  browsers.push('Android >= 4.4');
  browsers.push('iOS >= 10');
  browsers.push('ie 11');
}

module.exports = {
  browsers
};
```

raido:
PR https://github.com/tildeio/router.js/pull/292 will address this.

PR #292, opened 2020-01-18
Fix crashing with Babel polyfill / Chrome <= v37.0

raido:
Closes: https://github.com/tildeio/router.js/issues/291

It seems with Babel polyfill included `Object.isFrozen` vs `Object.isExtensible` has different meaning:

chrome-37-original-failure

This change was tested on:

* IE 10
* Chrome v27+
* Chrome latest
* IE 9 (has some other non related failure)

See screenshots below:

**IE 9:**
ie9

**IE 10:**
ie10

**Chrome v27**
chrome-27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 6
+  - 8
 
 sudo: false
 dist: trusty
--- a/lib/router/route-info.ts
+++ b/lib/router/route-info.ts
@@ -162,7 +162,7 @@ function createRouteInfoWithAttributes(
     },
   };
 
-  if (Object.isFrozen(routeInfo) || routeInfo.hasOwnProperty('attributes')) {
+  if (!Object.isExtensible(routeInfo) || routeInfo.hasOwnProperty('attributes')) {
     return Object.freeze(Object.assign({}, routeInfo, attributes));
   }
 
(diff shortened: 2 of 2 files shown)


raido:
Travis failed because transitive package seems to have dropped Node 6 support.

lib/router/utils.ts
```diff
@@ -12,6 +12,13 @@ export function isPromise<T>(p: any): p is Promise<T> {
   return p !== null && typeof p === 'object' && typeof p.then === 'function';
 }
 
+/**
+ * @private
+ */
+export function isFrozenOrNotExtensible(obj: object): boolean {
+  return Object.isFrozen(obj) || !Object.isExtensible(obj);
```
rwjblue: When is `!Object.isExtensible(obj)` not enough? Specifically, I'm wondering if `!Object.extensible(obj)` can ever be true for frozen objects?

raido: You are right. I changed it to just `!Object.isExtensible(obj)` and ran on same browsers (with Babel polyfill included) and tests pass.

I'll try to see if I can get Travis happy as well if not then I guess have to merge it as is?

raido:
I switched Travis from Node 6 -> 8.

We should drop Node <= 6 and release v7 of this package, after releasing v6.2.6 with this PR and backporting all the way to Ember.js v3.12 LTS over here - https://github.com/emberjs/ember.js/blob/master/package.json#L140

And then for Ember canary include v7 of this package.

raido:
Any updates on this?

raido:
@rwjblue ping :)

rwjblue:
Thank you @raido!

PR #293, opened 2020-07-17
Update queryParamsTransition to not change QueryParams if transition has been aborted

richgt:
https://github.com/emberjs/ember.js/issues/18568

### DIFF
--- a/tests/router_test.ts
+++ b/tests/router_test.ts
@@ -1500,6 +1500,62 @@ scenarios.forEach(function(scenario) {
       });
   });
 
+  test('abort query param only', function(assert) {
+    assert.expect(6);
+    map(assert, function(match) {
+      match('/').to('index');
+    });
+
+    routes = {
+      search: createHandler('search'),
+    };
+
+    let newParam = false;
+    let initial = true;
+
+    router.updateURL = function(updateUrl) {
+      url = updateUrl;
+      if (!initial) {
+        assert.ok(false, 'updateURL should not be called');
+      }
+    };
+
+    router.routeWillChange = (transition: Transition) => {
+      if (!transition.isAborted) {
+        if (newParam) {
+          assert.deepEqual(transition.to!.queryParams, { term: 'b' }, 'going to page with qps');
+          assert.deepEqual(
+            isPresent(transition.from) && transition.from!.queryParams,
+            {},
+            'from never has qps'
+          );
+        } else {
+          assert.strictEqual(transition.from, null, 'null transition from');
+          assert.deepEqual(transition.to!.queryParams, {}, 'empty transition queryParams');
+        }
+      }
+      if (!initial) {
+        if (!transition.isAborted) {
+          newParam = false;
+          transition.abort();
+        }
+      }
+    };
+
+    router.routeDidChange = (transition: Transition) => {
+      if (!transition.isAborted) {
+        assert.strictEqual(transition.from, null, 'routeDidChange null from transition');
+        assert.deepEqual(transition.to!.queryParams, {}, 'routeDidChange empty queryParams');
+      }
+    };
+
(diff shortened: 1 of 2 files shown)


richgt:
> Can you add a previously failing test (to make sure we don't regress in the future)?

Yup. In the plans, trying to work through figuring out if everything actually works first :-D.

rwjblue:
I think either a new test similar or a modification of this test would work:

https://github.com/tildeio/router.js/blob/master/tests/router_test.ts#L1416-L1501

tests/router_test.ts
```diff
@@ -1500,6 +1500,70 @@ scenarios.forEach(function(scenario) {
...
+      search: createHandler('search'),
+    };
+
+    let newParam = false;
+    let initial = true;
+
+    router.updateURL = function(updateUrl) {
+      url = updateUrl;
+      if (!initial) {
+        assert.ok(false, 'updateURL should not be called');
```
richgt, line 1519: @rwjblue - I was looking for something like `currentURL()`, but being in router.js, couldn't really find it, so this was the best I could come up with. If you have other ideas I'm certainly open. This does fail without the fix and pass with it.

tests/router_test.ts
```diff
@@ -1500,6 +1500,67 @@ scenarios.forEach(function(scenario) {
       });
   });
 
+  test('abort query param only', function(assert) {
+    assert.expect(6);
```
rwjblue, line 1504: There are 8 `assert.*` invocations (excluding the `assert.ok(false` one), why is this number 6? Is there an extra branch with assertions that we don't hit?

tests/router_test.ts
```diff
@@ -1500,6 +1500,67 @@ scenarios.forEach(function(scenario) {
...
+      if (!transition.isAborted) {
+        if (newParam) {
+          assert.deepEqual(transition.to!.queryParams, { term: 'b' }, 'going to page with qps');
+          assert.deepEqual(
+            isPresent(transition.from) && transition.from!.queryParams,
+            {},
+            'from never has qps'
+          );
+        } else {
+          assert.equal(transition.from, null);
```
rwjblue, line 1533: Can we `strictEqual` instead (since `equal` will coerce IIRC)?

```suggestion
          assert.strictEqual(transition.from, null);
```

tests/router_test.ts
```diff
@@ -1500,6 +1500,67 @@ scenarios.forEach(function(scenario) {
...
+      }
+    };
+
+    router.routeDidChange = (transition: Transition) => {
+      if (!transition.isAborted) {
+        if (newParam) {
+          assert.deepEqual(transition.to!.queryParams, { term: 'b' });
+          assert.deepEqual(isPresent(transition.from) && transition.from!.queryParams, {});
+        } else {
+          assert.equal(transition.from, null);
```
rwjblue, line 1551:
```suggestion
          assert.strictEqual(transition.from, null);
```

tests/router_test.ts
rwjblue: @richgt - As you resolve this, can you add assertion messages for each of the assertions that run (it will help figure out which are *not* running).

richgt: yeah, got rid of the one path that wasn't hit.

richgt: added assertion messages

tests/router_test.ts
```diff
@@ -38,50 +38,50 @@ let scenarios = [
   {
     name: 'Sync Get Handler',
     async: false,
-    getRoute: function(name: string) {
+    getRoute: function (name: string) {
```
rwjblue, line 41: Lots of unrelated changes here (by the looks of them they are prettier@2 related), can you roll them back?

richgt: Done

PR #297, opened 2020-07-20
Bump dependnecies

xg-wang:
Broccoli typescript compiler bump to bring typescript from 3.5 to 3.9.
Bump some broccoli deps that dropped node 8.


xg-wang:
broccoli-babel-transpiler isn't a straight forward bump, I also question the [compile targets](https://github.com/tildeio/router.js/blob/v6.2.6/ember-cli-build.js#L33-L58) 😛

rwjblue:
Thanks @xg-wang!

PR #304, opened 2020-08-24
feat(Transition): expose `previousTransition`

buschtoens:
image

### DIFF
--- a/lib/router/transition.ts
+++ b/lib/router/transition.ts
@@ -64,6 +64,7 @@ export default class Transition<T extends Route> implements Partial<Promise<T>> 
   isCausedByInitialTransition = false;
   isCausedByAbortingReplaceTransition = false;
   _visibleQueryParams: Dict<unknown> = {};
+  previousTransition: Maybe<Transition<T>>;
 
   constructor(
     router: Router<T>,
@@ -85,6 +86,7 @@ export default class Transition<T extends Route> implements Partial<Promise<T>> 
     this.targetName = undefined;
     this.pivotHandler = undefined;
     this.sequence = -1;
+    this.previousTransition = previousTransition;
 
     if (error) {
       this.promise = Promise.reject(error);


stefanpenner:
@buschtoens whats the motivation here?

buschtoens:
For the life of me, I cannot remember what I wanted this for. 😅

I _think_ the issue was that a new transition started by a redirect has the route that initiated the redirect as `from` and there is no way to get back to the original `from` that started the whole transition chain.

I think another issue might have been: before leaving / entering a specific route the user is redirected to an interstitial route. On this route there should be a button to retry the original transition.

rwjblue:
> I think the issue was that a new transition started by a redirect has the route that initiated the redirect as from and there is no way to get back to the original from that started the whole transition chain.

You can keep track of them with the router service fairly easily if that sort of thing is what you need, but it isn't obvious to me what the expected behavior here actually should be.

---

From [the router service RFC's section on routeDidChange](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#new-events-routewillchange--routedidchange):

> Redirection example:
> 
> 1. current route is A
> 1. user initiates a transition to B
> 1. routeWillChange fires from A to B.
> 1. B redirects to C
> 1. routeWillChange fires from A to C.
> 1. routeDidChange fires from A to C.

From [the router service RFCs section on `transition.from`](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#new-properties-from-and-to):

> I'm proposing we add from and to properties on transition whose values are RouteInfo instances representing the initial and final leafmost routes for this transition. Like all RouteInfos, these are read-only and internally immutable. They are not observable, because a transition instance is never changed after creation.

---

Reading through that, it is clear that you _can_ use the router services `routeWillChange` (keeping an array of the transition objects that happen from the first `routeWillChange` and the final `routeDidChange` which would reset the array).

However, it is not clear to me what "initial" in the `transition.from` section means. In other words, is the current behavior (that makes `.from` the target at the time of the redirect) wrong?

wagenet:
@buschtoens @rwjblue status?

PR #305, opened 2020-09-09
Add better Transition debugging information.

rwjblue:
Expose a few nice debugging tools when working with a transition:

```ts
interface Transition {
  /**
    In non-production builds, this function will return the stack that this Transition was
    created within. In production builds, this function will not be present.

    @method debugCreationStack
    @return string
  */
  debugCreationStack?: () => string | undefined;

  /**
    In non-production builds, this function will return the stack that this Transition was
    aborted within (or `undefined` if the Transition has not been aborted yet). In production
    builds, this function will not be present.

    @method debugAbortStack
    @return string
  */
  debugAbortStack?: () => string | undefined;

  /**
    In non-production builds, this property references the Transition that _this_ Transition
    was derived from or `undefined` if this transition did not derive from another. In
    production builds, this property will not be present.

    @property debugPreviousTransition
    @type {Transition | undefined}
  */
  debugPreviousTransition: Maybe<Transition<T>>;
}
```

---

Note: I've confirmed that Ember's build pipeline properly replaces the
`if (DEBUG) {` bits with `true` and `false` as appropriate:

  ```js
  if (true
      /* DEBUG */
  ) {
    var _error = new Error(`Transition creation stack`);

    this.debugCreationStack = () => _error.stack;

    this.debugPreviousTransition = previousTransition;
  }
```

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -37,12 +37,27 @@ function resolveRelativeModulePath(name, child) {
 resolveRelativeModulePath.baseDir = () => __dirname;
 
 function toAMD(tree) {
+  const isProduction = process.env.EMBER_ENV === 'production';
+  const isDebug = !isProduction;
+
   return new Babel(tree, {
     moduleIds: true,
     getModuleId: getRelativeModulePath,
     plugins: [
       ['module-resolver', { resolvePath: resolveRelativeModulePath }],
       ['@babel/plugin-transform-modules-amd', { noInterop: true }],
+      [
+        'babel-plugin-debug-macros',
+        {
+          flags: [
+            {
+              source: '@glimmer/env',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ember-cli-build.js [hunk 1]
```diff
@@ -37,12 +37,27 @@ function resolveRelativeModulePath(name, child) {
...
+        {
+          flags: [
+            {
+              source: '@glimmer/env',
+              flags: { DEBUG: isDebug, CI: !!process.env.CI },
+            },
+          ],
+        },
+        '@glimmer/env inlining',
+      ],
```
rwjblue, line 60: FWIW, this only affects the local tests (not what consumers use). Ember's build pipeline (as well as general Ember apps) uses nearly the same thing...

PR #306, opened 2020-09-10
Simplify TransitionState resolution system.

rwjblue:
* Reduce extraneous promise creation
* Remove `Continuation` system (use `transition.isAborted` directly)

### DIFF
--- a/lib/router/route-info.ts
+++ b/lib/router/route-info.ts
@@ -40,8 +40,6 @@ export interface Route {
   buildRouteInfoMetadata?(): unknown;
 }
 
-export type Continuation = () => boolean;
-
 export interface RouteInfo {
   readonly name: string;
   readonly parent: RouteInfo | RouteInfoWithAttributes | null;
@@ -222,16 +220,13 @@ export default class InternalRouteInfo<T extends Route> {
     return this.params || {};
   }
 
-  resolve(
-    shouldContinue: Continuation,
-    transition: InternalTransition<T>
-  ): Promise<ResolvedRouteInfo<T>> {
+  resolve(transition: InternalTransition<T>): Promise<ResolvedRouteInfo<T>> {
     return Promise.resolve(this.routePromise)
-      .then((route: Route) => this.checkForAbort(shouldContinue, route))
+      .then((route: Route) => this.checkForAbort(transition, route))
       .then(() => this.runBeforeModelHook(transition))
-      .then(() => this.checkForAbort(shouldContinue, null))
+      .then(() => this.checkForAbort(transition, null))
       .then(() => this.getModel(transition))
-      .then((resolvedModel) => this.checkForAbort(shouldContinue, resolvedModel))
+      .then((resolvedModel) => this.checkForAbort(transition, resolvedModel))
       .then((resolvedModel) => this.runAfterModelHook(transition, resolvedModel))
       .then((resolvedModel) => this.becomeResolved(transition, resolvedModel));
   }
@@ -375,8 +370,10 @@ export default class InternalRouteInfo<T extends Route> {
     });
   }
 
-  private checkForAbort<T>(shouldContinue: Continuation, value: T) {
-    shouldContinue();
+  private checkForAbort<U>(transition: InternalTransition<T>, value: U) {
+    if (transition.isAborted) {
+      throw new Error('Transition aborted');
+    }
 
     return value;
   }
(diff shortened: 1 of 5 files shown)


lib/router/route-info.ts
```diff
@@ -376,11 +376,9 @@ export default class InternalRouteInfo<T extends Route> {
   }
 
   private checkForAbort<T>(shouldContinue: Continuation, value: T) {
-    return Promise.resolve(shouldContinue()).then(function () {
-      // We don't care about shouldContinue's resolve value;
-      // pass along the original value passed to this fn.
-      return value;
-    }, null);
+    shouldContinue();
```
rwjblue, line 379: This either returns `true` or throws

rwjblue:
Note: I confirmed that Ember's test suite passes with these changes (when linked locally).

stefanpenner:
rebased on top of master + tested

lib/router/route-info.ts
```diff
@@ -40,7 +40,7 @@ export interface Route {
   buildRouteInfoMetadata?(): unknown;
 }
 
-export type Continuation = () => PromiseLike<boolean> | boolean;
+export type Continuation = () => boolean;
```
stefanpenner: Is it fair to say this new type is actually `() => true | never` ?

lib/router/transition-state.ts
```diff
@@ -42,40 +42,41 @@ export default class TransitionState<T extends Route> {
...
+      if (shouldContinue()) {
+        return true;
+      } else {
         // We distinguish between errors that occurred
         // during resolution (e.g. before"Model/model/afterModel),
         // and aborts due to a rejecting promise from shouldContinue().
         wasAborted = true;
-        return Promise.reject(reason);
-      }, currentState.promiseLabel('Handle abort'));
+        throw new Error('Transition aborted');
```
stefanpenner: given https://github.com/tildeio/router.js/pull/306/commits/e83c782d88d45c41cdfc5f6d1b23d74e15ad3481#r518934850 is this else unreachable?

tests/route_info_test.ts
```diff
@@ -43,9 +39,10 @@ test('RouteInfo can be aborted mid-resolve', function (assert) {
 
   let routeInfo = createHandlerInfo('stub');
 
-  function abortResolve() {
+  function abortResolve(): boolean {
```
stefanpenner: the return type should likely be `never`

lib/router/route-info.ts
stefanpenner: guess this is ultimately removed, so it doesn't matter.

lib/router/transition-state.ts
stefanpenner: guess this is ultimately removed, so it doesn't matter.

tests/route_info_test.ts
stefanpenner: guess this is ultimately removed, so it doesn't matter.

lib/router/route-info.ts
```diff
@@ -375,12 +370,12 @@ export default class InternalRouteInfo<T extends Route> {
     });
   }
 
-  private checkForAbort<T>(shouldContinue: Continuation, value: T) {
-    return Promise.resolve(shouldContinue()).then(function () {
-      // We don't care about shouldContinue's resolve value;
-      // pass along the original value passed to this fn.
-      return value;
-    }, null);
+  private checkForAbort<U>(transition: InternalTransition<T>, value: U) {
```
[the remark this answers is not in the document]
stefanpenner: actually, maybe something like the following signature should tighten things up nicely, as it will now be able to perform inference through `checkForAbort`

```ts
interface AbortableTransition<T extends boolean, R extends Route> extends InternalTransition<R> {
  isAborted: T;
}

function checkForAbort<U, T extends AbortableTransition<I, R>, R extends Route, I extends boolean>(
  transition: T,
  value: U
): I extends true ? never : U;
function checkForAbort<U, T extends AbortableTransition<I, R>, R extends Route, I extends boolean>(
  transition: T,
  value: U
): never | U {
  if (transition.isAborted) {
    throw new Error('Transition aborted');
  }

  return value;
}
```

That ended up being a little messier then expected, but now it's written I may as well add it.

stefanpenner (changes requested):

btecu:
@stefanpenner I was following the work here and noticed some changes that might be related to this pull request:
https://github.com/emberjs/ember.js/pull/16716

Do you think the issue is in `Ember.js` or could it be in `router.js`?

lib/router/route-info.ts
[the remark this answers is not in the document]
rwjblue: Nice, thanks for doing that @stefanpenner!

PR #307, opened 2020-10-29
fix: qp-only transition during initial transition

buschtoens:
Seems to fix https://github.com/emberjs/ember.js/issues/18577.

image

### DIFF
--- a/lib/router/router.ts
+++ b/lib/router/router.ts
@@ -120,7 +120,7 @@ export default abstract class Router<T extends Route> {
       // perform a URL update at the end. This gives
       // the user the ability to set the url update
       // method (default is replaceState).
-      let newTransition = new InternalTransition(this, undefined, undefined);
+      let newTransition = new InternalTransition(this, undefined, newState);
       newTransition.queryParamsOnly = true;
 
       oldState.queryParams = this.finalizeQueryParamChange(


jelhan:
Please see my comment here: https://github.com/emberjs/ember.js/issues/18577#issuecomment-719001736 Not sure if this is another bug or if this fix does not address the root cause.

lib/router/router.ts
```diff
@@ -120,7 +120,7 @@ export default abstract class Router<T extends Route> {
       // perform a URL update at the end. This gives
       // the user the ability to set the url update
       // method (default is replaceState).
-      let newTransition = new InternalTransition(this, undefined, undefined);
+      let newTransition = new InternalTransition(this, undefined, newState);
       newTransition.queryParamsOnly = true;
```
buschtoens, line 124:
```suggestion
      newTransition.queryParamsOnly = true;
      this.setupContexts(newState, newTransition);
```

Some other tests started breaking with the `undefined` -> `newState` change in L123, because this apparently leaks through Routes which haven't yet been passed through `setupContexts(...)` and thus did not have their `setup(...)` hook called.

Ultimately this causes [`Route#finalizeQueryParamChange`](https://github.com/emberjs/ember.js/blob/8f56dc80620a47aecb602e9f1ca18a1c85d83b9a/packages/%40ember/-internals/routing/lib/system/route.ts#L2547) to fail, as it [expects `route.controller` to be defined](https://github.com/emberjs/ember.js/blob/8f56dc80620a47aecb602e9f1ca18a1c85d83b9a/packages/%40ember/-internals/routing/lib/system/route.ts#L2572).

rwjblue:
This fix looks good conceptually, but we need to fixup the tests and add a specific test for the scenario described in https://github.com/emberjs/ember.js/issues/18577

Mikek2252:
@buschtoens @rwjblue Is this solution still good? if so is there anything i can do to help get this merged? Happy to write the test for it if i can be pushed in the correct general direction, im not overly familiar with this code.

vstefanovic97:
Hey, I'm still seeing this problem in 2025, anything I can do to help?

PR #310, opened 2020-11-06
Add failing test for native promises that return another transition.

rwjblue:
Failing test for `TransitionAborted` via `router.transitionTo` within async hook.

Builds off of #306.


wagenet:
@rwjblue status?

rwjblue:
This is still an issue that needs to be fixed (these failing tests should be made to pass).

PR #311, opened 2020-11-06
Upgrade CI

stefanpenner:
* use volta
* test across multiple node versions and OS's
* run tests weekly on sundays to catch odd things

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,19 +2,33 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+    branches:
+      - master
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,19 +2,33 @@ name: CI
...
-  build:
-    runs-on: ubuntu-latest
+  test:
+    name: Tests
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        node: ['10', '12', '14', '15']
+        os: [ubuntu-latest, macOS-latest, windows-latest]
```
rwjblue, line 22: FWIW, I don't think this provides any value. All of our tests run in the browser which is the same across all of these matrix setups...

PR #319, opened 2021-03-04
Ensure query params are preserved through an intermediate loading state transition

sly7-7:
related to https://github.com/emberjs/ember.js/pull/19448

Previously, the assertion on line 5400 would have failed

### DIFF
--- a/lib/router/transition-intent/named-transition-intent.ts
+++ b/lib/router/transition-intent/named-transition-intent.ts
@@ -140,7 +140,7 @@ export default class NamedTransitionIntent<T extends Route> extends TransitionIn
     }
 
     merge(newState.queryParams, this.queryParams || {});
-    if (isIntermediate) {
+    if (isIntermediate && oldState.queryParams) {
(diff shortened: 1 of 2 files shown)


lib/router/transition-intent/named-transition-intent.ts
```diff
@@ -140,6 +140,9 @@ export default class NamedTransitionIntent<T extends Route> extends TransitionIn
     }
 
     merge(newState.queryParams, this.queryParams || {});
+    if (isIntermediate) {
+      merge(newState.queryParams, oldState.queryParams);
+    }
```
rwjblue, line 145: I think you'll have to either default like this:

```js
    if (isIntermediate) {
      merge(newState.queryParams, oldState.queryParams || {});
    }
```

Or only do it if there are QPs (seems vaguely preferable):

```suggestion
    if (isIntermediate && oldState.queryParams) {
      merge(newState.queryParams, oldState.queryParams);
    }
```

PR #321, opened 2021-03-07
introduce isIntermediate flag in transition

sly7-7:
This is a prerequisite to fix https://github.com/emberjs/ember.js/issues/19449

cc/ @rwjblue


rwjblue (commented):
Can you update the internal transition constructor to set `this.isIntermediateTransition = false`?

rwjblue (approved):
Never mind! I misread the diff on mobile. 😝

PR #322, opened 2021-03-09
delete context during setup only if route is defined

sly7-7:
Necessary for https://github.com/emberjs/ember.js/pull/19266

@rwjblue Do you think there could be a demonstrating failing test in this repo ?


wagenet:
@sly7-7 status?

PR #324, opened 2021-05-16
README: change getHandler to getRoute

### LINKED ISSUE #264: Is README.md correct for defining handler lookup function?
The README.md file says:

```
  router.getHandler = function(name) {
    return myHandlers[name];
  };
```

This doesn't seem to do anything, but having dug into the library it seems to error on `this.router.getRoute()` not being defined. Should the README.md say that `router.getRoute` needs to be set instead?

With that said, I did change my code to set getRoute, which did stop the error, but my handler is still not being called. Is there a working example anywhere that I could follow?

Quasar-Kim:
This PR changes getHandler method in README to getRoute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ myHandlers.newPost = {
   }
 };
 
-router.getHandler = function(name) {
+router.getRoute = function(name) {
   return myHandlers[name];
 };
 ```


wagenet:
@Quasar-Kim thanks!

PR #328, opened 2022-01-03
Reset query params after aborting transition in RouterService#routeWillChange

akluball:
Reset the query params to their original values after aborting a transition in the RouterService#routeWillChange hook.

Consider a transition that changes a query param and refreshModel is true for that query param. If, the transition is aborted in the RouterService#routeWillChange hook, an infinite cycle is created.

Currently, at the end of [router#getTransitionByIntent](https://github.com/tildeio/router.js/blob/4f5d411e9ba44efea003bc3159484b4060ebfc2a/lib/router/router.ts#L278), the `fireQueryParamDidChange` method is always called with the changed query parameters without consideration for whether the transition was aborted. The `fireQueryParamDidChange` method triggers a 'queryParamsDidChange' event which is handled in the [Ember route class](https://github.com/emberjs/ember.js/blob/e317a882cd6d0950810a621de6d9154c7d75560e/packages/%40ember/-internals/routing/lib/system/route.ts#L2364) and triggers a refresh if `refreshModel` is true for the changed query param. This forms a `getTransitionByIntent` > `fireQueryParamDidChange` > `refresh` > `getTransitionByIntent` infinite cycle.

Simply skipping the `fireQueryParamDidChange` call when the transition is aborted is not sufficient. In the query param only transition flow, the query param on the controller still gets updated based on how that flow is structured.

The proposed solution resets the query params when the transition is aborted by calling fireQueryParamDidChange with the original query params as end state.

### DIFF
--- a/lib/router/router.ts
+++ b/lib/router/router.ts
@@ -275,6 +275,9 @@ export default abstract class Router<T extends Route> {
       this.notifyExistingHandlers(newState, newTransition);
     }
 
+    if (newTransition.isAborted) {
+      queryParamChangelist = getChangelist(newState.queryParams, oldState!.queryParams);
+    }
     this.fireQueryParamDidChange(newState, queryParamChangelist!);
 
     return newTransition;
--- a/tests/query_params_test.ts
+++ b/tests/query_params_test.ts
@@ -13,7 +13,11 @@ import {
   trigger,
 } from './test_helpers';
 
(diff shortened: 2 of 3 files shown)


wagenet:
@akluball are you using this as part of Ember? If so, maybe also file an issue on https://github.com/emberjs/ember.js for quicker handling.

PR #339, opened 2024-03-09
Fix a type error when `exactOptionalPropertyTypes` is enabled

boris-petrov:
Fixes https://github.com/emberjs/ember.js/issues/20500

cc @wagenet


wagenet:
It seems like there would be a lot of other things affected by this, including the line immediately above: `queryParams?: Dict<unknown>;`

boris-petrov:
Right, this definitely doesn't fix all issues in this repo concerning `exactOptionalPropertyTypes` but it does fix all issues in a user's Ember app when using `exactOptionalPropertyTypes`.

PR #341, opened 2026-01-20
allow to pass data to the transition before routeWillChange is fired

frederikbosch:
When calling `.transitionTo()` it is not possible to pass `data` to the transition before `routeWillChange` is fired. The called method returns after `routeWillChange` was already fired. The PR allows to pass `data` with `transitionTo()` so it is available to events listening for `routeWillChange`.


frederikbosch:
I will drop this PR as I now know people are working on a successor of this router.

PR #342, opened 2026-01-20
add Transition.waitFor() method

frederikbosch:
The method `Transition.waitFor()` allows to wait for another promise to finish before continuing the transition into the new route. In particularly, if you want to add a view transition to the route change, this could look like this when this PR gets accepted.

```js
router.on('routeWillChange', async (transition) => {
  const { promise, resolve, reject } = Promise.withResolvers();
  transition.waitFor(promise);

  const viewTransition = document.startViewTransition(async () => {
    resolve();
    await transition.promise;
  });
  await viewTransition.updateCallbackDone;
});
```


frederikbosch:
You can now also wait for query param changes, allowing those transitions to be part of a view transition too.

kategengler:
Since this would add public API to Ember, it requires an [RFC](https://github.com/emberjs/rfcs#when-you-need-to-follow-this-process)

frederikbosch:
Thanks @kategengler, I was not aware. I will follow up.

featheredtoast:
Has there been any followup on this/RFC proposed?

I was also expecting a promise to be handled in routeWillChange, something like:

```
router.on('routeWillChange', (transition) => {
  return document.startViewTransition(() => transition.promise).ready;
});
```

Route change syncs on the view transition's ready promise, view transition's callback awaits the transition's promise.

frederikbosch:
Just added an [RFC for a new Transition.waitFor method](https://github.com/emberjs/rfcs/pull/1183).

